PR #243, opened 2015-02-01
updated deprecated webdriverjs to webdriverio

christian-bromann:
Hi,

I updated your webdriverjs version. We are calling it webdriverio now to avoid namespace confusions with the selenium-webdriver project. I haven't tried to run these tests but webdriverio should be backwards compatible.


lexoyo:
That's cool, thx :zap:

PR #286, opened 2015-07-04
removes npm-debug.log from git

pierreozoux:
To avoid commiting npm errors to git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .temp
 .cache
 node_modules
+npm-debug.log
 dist/server/sessions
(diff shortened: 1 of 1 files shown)


lexoyo:
:+1: 
thx
PS: big fan of indiehosters here :)

pierreozoux:
Oho :) Nice to hear! The dockerfile is comming with support for IndieHosters :)

simison:
BTW, `*.log` is even more bullet proof. ;-)

pierreozoux:
@simison good catch :)

PR #288, opened 2015-07-04
Dockerfile

pierreozoux:
You can use your own automated build then to remove dependencie over IndieHoster automated build from the readme.

Best!


lexoyo:
Even nicer, thx!
:+1:

PR #291, opened 2015-07-06
Fixes fat docker image

pierreozoux:
In a Dockerfile, if you don't do this `rm` thing,  the image stays with all apt tree. This commit makes the image slimmer.

(Sorry I should have put that from the beginning)


lexoyo:
Ok cool
I did not know that

PR #383, opened 2016-02-03
Use the right namespace

pierreozoux:
tell me if oyu want to be added to the organisation silexlabs on dockerhub.


lexoyo:
cool :+1: 

> tell me if oyu want to be added to the organisation silexlabs on dockerhub.

Sure! I'm lexoyo there too :)

PR #402, opened 2016-04-23
Couchcms

Fireheadii:


lexoyo:
hi @Fireheadii 

the couchcms branch is a test I did to create a couchcms template with Silex, but I never really knew if it is useful...

you created this "pull request", which means you would like to merge the changes from the "couchcms" branch into the "master" branch. is this what you intended to do?

Issue #424, opened 2016-05-23
Edit online then publish = css broken ?


Draky50110:
Hello.

Before editing editable.html : http://www.draky.net/silex_cv_avant/

Then edit in editor.silex.org changing just some texts then publish : http://www.draky.net/silex_cv_apres/

Something's broken...

lexoyo:
Hi @Draky50110 

I am sorry this is a bug on my side: when you publish on FTP it corrupts several scripts. This is the same issue as this one (in French)
https://github.com/silexlabs/Silex/issues/421

I will definitely fix it, but meanwhile the possible workarounds:
* you can replace these scripts "by hand' with an FTP client, just download them and put them with the same name in the `js` folder: [jquery.js](http://www.silex.me/js/jquery.js), [jquery-ui.js](http://www.silex.me/js/jquery-ui.js), [pageable.js](http://www.silex.me/js/pageable.js), [front-end.js](http://www.silex.me/js/front-end.js), and the same for these CSS which should go in the folder `css` [front-end.css](http://www.silex.me/css/front-end.css), [normalize.css](http://www.silex.me/css/normalize.css)
* or you can publish to a folder in Dropbox and transfer all your site to FTP by hand (or there are tools which can sync dropbox and ftp)

Sorry about that, it will take a while to fix it but I will eventually - or maybe someone will help with this bug which is probably in [silex-tasks.js](https://github.com/silexlabs/Silex/blob/develop/dist/server/silex-tasks.js#L103) which uses [unifile](https://github.com/silexlabs/unifile/) cc @JbIPS

lexoyo:
hello
I have tried to fix the issue (https://github.com/silexlabs/Silex/pull/425)
could you try again to publish on FTP directly please?
do not forget to clear your browser cache
cc @fecou

Draky50110:
Seems etter but image (CV picture) does not seems to be uploaded or correctly linked as it does not appear in the published website.
http://www.draky.net/silex_cv_apres/

lexoyo:
ok cool 
can you double click on the image in Silex to replace it with the same image and publish again?

fecou:
For me it still doesn't work with the images (cache cleared and images changed). All the rest works well.

lexoyo:
Hi, I changed something again, could you confirm it work with both dropbox and ftp?
I am dealing with an old code in unifile, which is being refactored by @JbIPS (I can't wait to have the new version ;)

fecou:
[...]

PR #425, opened 2016-05-25
attempt to fix ftp publication issue #424

lexoyo:


fecou:
Hello. It works with all the js and css files that didn't work before (see #424). But know, the script.js and the images are not published.

PR #518, opened 2016-12-02
Add Cloud Foundry Integration

yacloud:
Add an option to host your own Silex site in [Bluemix](https://www.ng.bluemix.net), as for now , Bluemix or CloudFoundry only support one runtime in one app, so we need to update the build phase not using java but only using nodejs to build the assets. And the PR only illustrate publishing the site to Bluemix, it will handle all app related staff, building & running & monitoring...

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "scripts": {
     "start": "pm2 start .pm2.json",
     "start:debug": "SILEX_DEBUG=true node dist/server/server.js",
-    "build": "npm run build:css && npm run build:html && npm run build:js:release ",
+    "build": "npm run build:css && npm run build:html",
     "build:css": "node_modules/.bin/lessc src/css/_styles.less dist/client/css/admin.css --source-map",
     "build:html": "node_modules/.bin/jade src/html/index.jade --out dist/client/ --no-debug",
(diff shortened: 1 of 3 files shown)


yacloud:
maybe we can merge it to another branch

package.json
```diff
@@ -19,7 +19,7 @@
   "scripts": {
     "start": "pm2 start .pm2.json",
     "start:debug": "SILEX_DEBUG=true node dist/server/server.js",
-    "build": "npm run build:css && npm run build:html && npm run build:js:release ",
```
lexoyo: if you remove this, the file `dist/client/js/admin.js` will not be generated, so Silex will not work at all

lexoyo:
> we need to update the build phase not using java but only using nodejs to build the assets

It would be nice to use another compiler than google closure, which is very efficient but not very trendy anymore - and in java ;) 

We could use babel instead

For this we need to replace google closure module management (`goog.provide` and `goog.require`) by standard modules management such as requirejs

lexoyo:
Also the file `README2CF.md ` should go to [Silex wiki](https://github.com/silexlabs/Silex/wiki), can you move it there?

package.json
yacloud: i don't know much more about the Silex architecture, i thought it just build the final release, is that right?After remove the java build phase, i deployed the site successfully, maybe some functionality not works well, is there any function works using the `admin.js` ? For now, the simple usage works fine, here the [site](http://yadesigner.au-syd.mybluemix.net/) i deployed, would mind verify that

yacloud: yeah, so strange..., i can't remember exactly how its built, but anyway do you have any plan to migrate to babel?

lexoyo: not really
unless someone is ready to spend time doing it ;)

lexoyo: another solution is to build silex on a CI server (travis or codeship...) and deploy to bluemix the compiled files...

yacloud: Okay, I'll take that workaround and update the readme then move it to the wiki

lexoyo: thank you @yacloud !

PR #583, opened 2017-04-10
handle the right click

tariqbenezza:
Too many people using Silex tried to right click on the workspace.

I propose to help them by handle this right click and display a message :

image

Tell me if it's Ok for you ?

Thank you in advance

### DIFF
--- a/src/css/stage.less
+++ b/src/css/stage.less
@@ -46,27 +46,37 @@
   }
 }
 
-#powDiv {
+/**
+* Right click button style
+**/
+
+.rightClick {
     position: absolute;
     width: 150px;
     text-align: center;
-    background-color: #252525;
+    background-color: @silexdarkgrey;
     color: #FFF;
     width: 200px;
     height: 46px;
     border-radius: .15em;
     border: 1px solid #444;
     z-index: 11111;
+    top:20000px;
+    left:20000px;
 }
-#leftPowDiv {
+.parentImgRightClick {
     display: inline;
 }
-#imgLeftPowDiv {
+.imgRightClick {
     width: 70px;
     height: 38px;
     margin-top: 5px;
+    float:left;
 }
-#rightPowA {
+.textRightClick {
     display: inline;
-    color: #777
+    color: @silexlightergrey;
+    position: relative;
+    top: 25%;
+    font-size: 13px;
 }
--- a/src/css/stage.less
+++ b/src/css/stage.less
@@ -45,3 +45,28 @@
     background-color: @silexlightgrey;
   }
 }
+
+#powDiv {
+    position: absolute;
+    width: 150px;
+    text-align: center;
+    background-color: #252525;
+    color: #FFF;
+    width: 200px;
+    height: 46px;
+    border-radius: .15em;
+    border: 1px solid #444;
+    z-index: 11111;
+}
+#leftPowDiv {
+    display: inline;
+}
+#imgLeftPowDiv {
+    width: 70px;
+    height: 38px;
+    margin-top: 5px;
+}
+#rightPowA {
+    display: inline;
+    color: #777
+}
--- a/src/js/view/context-menu.js
+++ b/src/js/view/context-menu.js
@@ -59,7 +59,13 @@ silex.view.ContextMenu = function(element, model, controller) {
 
 };
 
+/**
+ * Get the silex context menu
+ * @static 
+ * @type {string}
+ */
 silex.view.ContextMenu.CLASS_NAME='silex-context-menu';
+
 /**
  * create the context menu
  * called by the app constructor
--- a/src/js/view/context-menu.js
+++ b/src/js/view/context-menu.js
@@ -59,7 +59,7 @@ silex.view.ContextMenu = function(element, model, controller) {
 
 };
 
-
+silex.view.ContextMenu.CLASS_NAME='silex-context-menu';
 /**
  * create the context menu
  * called by the app constructor
--- a/src/js/view/menu.js
+++ b/src/js/view/menu.js
@@ -55,7 +55,7 @@ silex.view.Menu = function(element, model, controller) {
   this.controller = controller;
 };
 
-
+silex.view.Menu.CLASS_NAME='menu-container';
 /**
  * reference to the menu class of the closure library
  */
--- a/src/js/view/menu.js
+++ b/src/js/view/menu.js
@@ -55,7 +55,13 @@ silex.view.Menu = function(element, model, controller) {
   this.controller = controller;
 };
 
+/**
+ * Get the menu container
+ * @static 
+ * @type {string}
+ */
 silex.view.Menu.CLASS_NAME='menu-container';
+
 /**
  * reference to the menu class of the closure library
  */
--- a/src/js/view/stage.js
+++ b/src/js/view/stage.js
@@ -305,6 +305,14 @@ silex.view.Stage.prototype.initEvents = function(contentWindow) {
   this.contentDocument = contentWindow.document;
   this.contentWindow = contentWindow;
 
+ //detect right click
+  goog.events.listen(contentWindow.document, 'contextmenu', function(event) {
+    let x = event.clientX;
+    let y = event.clientY;
+    this.handleRightClick(x, y);
+    event.preventDefault();
+    return false;
+  }, false, this);
   // listen on body instead of element because user can release
   // on the tool boxes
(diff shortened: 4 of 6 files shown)


src/css/stage.less
```diff
@@ -45,3 +45,28 @@
...
+}
+#imgLeftPowDiv {
+    width: 70px;
+    height: 38px;
+    margin-top: 5px;
+}
+#rightPowA {
+    display: inline;
+    color: #777
+}
```
lexoyo: why `pow` in the names?
please add a comment stating that this section is about right click

lexoyo: also, please use the existing less variables for colors, see `src/css/_variables.less`, e.g. `background-color: @silexmediumgrey;`

src/js/view/context-menu.js
```diff
@@ -59,7 +59,7 @@ silex.view.ContextMenu = function(element, model, controller) {
 
 };
 
-
+silex.view.ContextMenu.CLASS_NAME='silex-context-menu';
```
lexoyo: please add a comment with `@static` keyword and `@type {string}`

src/js/view/menu.js
```diff
@@ -55,7 +55,7 @@ silex.view.Menu = function(element, model, controller) {
   this.controller = controller;
 };
 
-
+silex.view.Menu.CLASS_NAME='menu-container';
```
lexoyo: please add a comment with `@static` keyword and `@type {string}`

src/js/view/stage.js
```diff
@@ -1212,3 +1220,44 @@ silex.view.Stage.prototype.moveElements = function(elements, offsetX, offsetY) {
 silex.view.Stage.prototype.isMobileMode = function() {
   return goog.dom.classlist.contains(document.body, 'mobile-mode');
 };
+
+silex.view.Stage.prototype.handleRightClick = function(x, y) {
```
lexoyo: please add comments with type info, something like
```
/**
 * @param {number} x
 * @param {number} y
 */

File: src/js/view/stage.js [hunk 5]
```diff
@@ -1212,3 +1220,44 @@ silex.view.Stage.prototype.moveElements = function(elements, offsetX, offsetY) {
...
+
+  if(document.getElementById( 'powDiv' )!=null){
+    var el = document.getElementById( 'powDiv' );
+    el.parentNode.removeChild( el );  
+  }
+
+  var powDiv=goog.dom.createElement('div');
+  powDiv.id='powDiv';
+  
+  var leftPowDiv=goog.dom.createElement('div');
```
lexoyo: this should be defined in the jade/pug/html templates, probably in `src/html/templates/workspace.jade`

src/js/view/stage.js
```diff
@@ -1212,3 +1220,44 @@ silex.view.Stage.prototype.moveElements = function(elements, offsetX, offsetY) {
...
+  powDiv.style.top = (y+myContextHeight)+'px';
+  
+  document.body.appendChild(powDiv);
+  document.getElementById('powDiv').appendChild(leftPowDiv);
+  document.getElementById('leftPowDiv').appendChild(imgLeftPowDiv);
+  document.getElementById('powDiv').appendChild(rightPowA);
+
+  setTimeout(function(){ 
+    var el = document.getElementById( 'powDiv' );
+    el.parentNode.removeChild( el );
```
lexoyo: you may want to check that `el.parentNode` is defined, since it won't if the user right clicks many times?

src/js/view/stage.js
```diff
@@ -1212,3 +1220,44 @@ silex.view.Stage.prototype.moveElements = function(elements, offsetX, offsetY) {
...
+    el.parentNode.removeChild( el );  
+  }
+
+  var powDiv=goog.dom.createElement('div');
+  powDiv.id='powDiv';
+  
+  var leftPowDiv=goog.dom.createElement('div');
+  leftPowDiv.id='leftPowDiv';
+
+  var imgLeftPowDiv=goog.dom.createElement('img');
```
lexoyo: same here, this should be defined in the html templates

src/js/view/stage.js
```diff
@@ -1212,3 +1220,44 @@ silex.view.Stage.prototype.moveElements = function(elements, offsetX, offsetY) {
...
+  imgLeftPowDiv.id='imgLeftPowDiv';
+  imgLeftPowDiv.setAttribute('src', '../assets/logo-silex.png');
+
+
+  var rightPowA=goog.dom.createElement('a');
```
lexoyo: same here, this should be defined in the html templates

src/js/view/stage.js
```diff
@@ -1212,3 +1220,44 @@ silex.view.Stage.prototype.moveElements = function(elements, offsetX, offsetY) {
...
+  rightPowA.id='rightPowA';
+  rightPowA.setAttribute('href', 'http://www.silex.me');
+  rightPowA.setAttribute('target', '_blank');
+  rightPowA.innerHTML='Powered by Silex';
+
+  var myContextHeight=goog.dom.getElementByClass(silex.view.ContextMenu.CLASS_NAME).offsetHeight;
+  var myMenuWidth=goog.dom.getElementByClass('menu-container').offsetWidth;
```
lexoyo: here you can use the constant you have created `silex.view.Menu.CLASS_NAME`

src/js/view/stage.js
```diff
@@ -1212,3 +1220,44 @@ silex.view.Stage.prototype.moveElements = function(elements, offsetX, offsetY) {
...
+  powDiv.style.left = (x+myMenuWidth)+'px';
+  powDiv.style.top = (y+myContextHeight)+'px';
+  
+  document.body.appendChild(powDiv);
+  document.getElementById('powDiv').appendChild(leftPowDiv);
+  document.getElementById('leftPowDiv').appendChild(imgLeftPowDiv);
+  document.getElementById('powDiv').appendChild(rightPowA);
+
+  setTimeout(function(){ 
+    var el = document.getElementById( 'powDiv' );
```
lexoyo: you can use the `rightPowA` variable here, it is defined in the parent scope

src/js/view/stage.js
```diff
@@ -1212,3 +1220,44 @@ silex.view.Stage.prototype.moveElements = function(elements, offsetX, offsetY) {
...
+
+  powDiv.style.left = (x+myMenuWidth)+'px';
+  powDiv.style.top = (y+myContextHeight)+'px';
+  
+  document.body.appendChild(powDiv);
+  document.getElementById('powDiv').appendChild(leftPowDiv);
+  document.getElementById('leftPowDiv').appendChild(imgLeftPowDiv);
+  document.getElementById('powDiv').appendChild(rightPowA);
+
+  setTimeout(function(){
```
lexoyo: please use the arrow function here (`setTimeout(() => { `)

src/js/view/stage.js
```diff
@@ -1212,3 +1220,44 @@ silex.view.Stage.prototype.moveElements = function(elements, offsetX, offsetY) {
...
 };
+
+silex.view.Stage.prototype.handleRightClick = function(x, y) {
+
+  if(document.getElementById( 'powDiv' )!=null){
+    var el = document.getElementById( 'powDiv' );
+    el.parentNode.removeChild( el );  
+  }
+
+  var powDiv=goog.dom.createElement('div');
```
lexoyo: when possible use `const` instead of `var`, or `let` if you intend to modify the value

src/css/stage.less
tariqbenezza: I used ```pow``` because of the popin wording "**Pow**ered by Silex"

src/js/view/context-menu.js
tariqbenezza: Is it OK for you if I do this : 

```
/**
* Get the silex context menu via a @static @type {string} field
*/
```

src/js/view/menu.js
tariqbenezza: Same as above. Is it OK for you if I write : 

```
/**
* Get the menu container via a @static @type {string} field
*/
```

src/js/view/context-menu.js
lexoyo: the "@" keywords need to be on different lines and at the start of the line
you can look at how it is in the other parts of the code

lexoyo: also do not forget the " " in front of the "*"

lexoyo:
Yeah! Thats what I'm talkin' about ;)
:+1:

PR #598, opened 2017-05-07
Fix markdown headers missing a space in readme.md

rapenne-s:
3 headers ## were missing a space between the hashtags and the text


lexoyo:
Good catch, thank you Solène !

Issue #624, opened 2017-06-29
Google analytics embed code breaks my site


antosilex:
All the sites i've done with silex since 2 years are broken even the last one done 3 months ago.
I'd like to modify the site http://www.clairiere-epona.com/ but here is what i see on my editor
![capture du 2017-06-29 14 53 59](https://user-images.githubusercontent.com/6757106/27688509-13965e04-5cdb-11e7-8685-1b3f75df3515.png)

lexoyo:
Hello
Does it display the "update warning" when you open the site?
If you click "preview", is the preview broken too?

antosilex:
Hi
Yes i saw the "update warning" and i saved as a new version but it doesn't change anything.
The preview is broken too
![capture du 2017-06-30 15 05 02](https://user-images.githubusercontent.com/6757106/27736760-db5c739a-5da5-11e7-85fb-cb384551ca30.png)

lexoyo:
Can you please share the website somewhere?

antosilex:
You already are the owner of the shared folder
I sent you an email
Tks

singchan:
There is an [issue with google analytics embed code in this method](https://github.com/silexlabs/Silex/blob/2d95ee981d25b56d4e966642fde21db6ba3b4f0a/src/js/utils/url.js#L217)

1- please remove this code from your sites `(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)})(window,document,'script','//www.google-analytics.com/analytics.js','ga');`
2- we should remove these crazy regexp

antosilex:
Tks for your answers
I've tried to remove the Google Analytics JS code  i'm not sure to do it properly  but it doesn't change anything, i really need some help to solve this problem it's urgent to update the website with the new planning for 2017, i don't know what to do all my assets disapeared and all the container are broken, and i really need to give some answers to the website owners, is there hope to save my previous work ?

lexoyo:
is this issue resolved @antosilex ?

PR #716, opened 2018-03-30
Server side conversions

lexoyo:
When Silex saves / opens / publishes a website, it parses the website's DOM to find URLs and make them absolute (when opening) or relative (when saving) or relative to the published version

This PR takes all these operations on the server side, taking advantage of nodejs parsing libraries, and making it possible to parse the DOM without loading any external images or scripts or styles or iframes, and in a cross browser consistent way

This is way cleaner (no more regexp to find the URLs for example), more efficient on the client side, less efficient on the server side, and it fixes long time bugs like #624 

Backward compatibility is dropped for this release because I did not manage to port it from client side to server side, but this version is 100% compatible with the one currently on master. So now when you try to open an older website, Silex sends you to https://old.silex.me/ which has the version currently on master. There you can open your website and save it with the version which still has BC on the client side. Then come back to editor.silex.me and you will be able to open it. A BC system is already ready on this PR but it is waiting for the next version to be used.


dist/server/BackwardCompat.js
```diff
@@ -0,0 +1,167 @@
...
+
+const Path = require('path');
+const constants = require('./Constants.json');
+
+/**
+ * the version of the website is stored in the generator tag as "Silex v-X-Y-Z"
+ * used for backward compat
+ * also the static files are taken from //{{host}}/static/Y-Z
+ */
+const LATEST_VERSION = require(Path.resolve(__dirname, '../../package.json')).version.split('.');
```
JbIPS: FYI, if you start the server with NPM, it sets an env variable for you : `process.env.npm_package_version`

dist/server/PublishJob.js
```diff
@@ -0,0 +1,413 @@
...
+      if(err.code != 'ENOENT') {
+        throw err;
+      }
+    });
+
+    // start by testing if the folders exist before creating them
+    // then download all assets
+    // FIXME: should use unifile's batch method to avoid conflicts or the "too many clients" error in FTP
+    //return Promise.all([
+    return sequential([
```
JbIPS: As previously discussed you should do that. But maybe you want to clean that in another PR

dist/server/PublishRouter.js
```diff
@@ -0,0 +1,61 @@
...
+
+module.exports = function(port, rootUrl, unifile) {
+
+  const router = express.Router();
+
+  // **
+  // publication tasks
+  router.post('/tasks/:task', bodyParser.json(), (req, res, next) => {
+    // init the session (shouldn't it be done by express??
+    req.session.sessionID = req.session.sessionID || uuid.v4();
```
JbIPS: Yes it should. Did you encounter a case where the session was not set?

dist/server/WebsiteRouter.js
```diff
@@ -0,0 +1,230 @@
...
+//
+// Silex is available under the GPL license
+// http://www.silexlabs.org/silex/silex-licensing/
+//////////////////////////////////////////////////
+
+const Path = require('path');
+const express = require('express');
+const fs = require('fs');
+const { JSDOM } = require('jsdom');
+const bodyParser = require('body-parser');
```
JbIPS: As seen in CE2, you don't need `body-parser` since `express 4.16.0`

dist/server/index.js
```diff
@@ -0,0 +1,68 @@
...
+  name: 'silex-session',
+  secret: process.env.SILEX_SESSION_SECRET || 'test session secret'
+}));
+
+// create the routes for unifile/CloudExplorer
+// and for Silex tasks
+const cloudExplorerRouter = new CloudExplorerRouter(rootUrl);
+app.use(new WebsiteRouter(port, rootUrl, cloudExplorerRouter.unifile));
+app.use(new PublishRouter(port, rootUrl, cloudExplorerRouter.unifile));
+app.use(new SslRouter(app));
```
JbIPS: I know it's arguable, but IMO this file should act as a controller and perform the environment checks (like SSL and such) before using the router.
In that case, the router wouldn't have to worry about the env and the configuration variables should be injected in its constructor. 

This will greatly simplify testing.

package.json
```diff
@@ -38,34 +40,33 @@
     "start:electron": "SILEX_ELECTRON=true electron ."
   },
   "dependencies": {
-    "body-parser": "1.13.3",
+    "bluebird": "*",
```
JbIPS: I don't think that's what you want. Even with a `package-lock` this file should be human readable (the lock clearly isn't).

Basically, we talked about express version and it's now very difficult to know which one is used here.

src/js/service/cloud-storage.js
```diff
@@ -66,76 +66,117 @@ silex.service.CloudStorage.prototype.ready = function(cbk) {
...
+  // .then(() => {
+  //   cbk();
+  // })
+  // .catch(e => {
+  //   console.error('Error: could not write file', fileInfo, e);
+  //   if (opt_errCbk) opt_errCbk(/** @type {Object} */ (e));
+  // });
+	const oReq = new XMLHttpRequest();
+	oReq.onload = function(event) {
+		if(oReq.status === 200) {
```
JbIPS: What happened to your indentation? (╯°□°）╯︵ ┻━┻

lexoyo:
Thanks, great review, I'll fix all this

dist/server/PublishJob.js
lexoyo: Yes as you told me it is better to test if a file exist and by writing and checking the error

But I would rather do that in an other PR because this current solution is the result of many problems with unifile batch and I am not comfortable changing it now

dist/server/PublishJob.js
```diff
@@ -0,0 +1,413 @@
...
+        name: 'writefile',
+        path: this.jsFile,
+        content: this.tree.scriptTags.reduce((prev, tag) => prev + '\n' + tag.innerHTML, ''),
+      });
+    }
+    const batchActionsWithAssets = batchActions.concat(
+      assets
+      .filter(file => !!file)
+      .map(file => {
+        if(file.path === 'css/') {
```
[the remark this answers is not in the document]
lexoyo: oops old code sorry

dist/server/PublishRouter.js
lexoyo: I guess I did
I'll remove it and see what happens since we updated express

dist/server/WebsiteRouter.js
lexoyo: Concerning JSON: CE adds routes that collide with Silex new routes, e.g. `/website/github/get/master/` will be caught by CE as a service called "website/github". So I need to put Silex router before CE router, and so the body needs to be explicit

Concerning text: here I need the body parsed as text, and I can't figure out how to access the body parser from express to use bodyParser.text(

Also in the docs, this is still written to require('body-parser') so I lost too much time with this for now
https://expressjs.com/en/4x/api.html#req.body

package.json
lexoyo: i agree but one of the dependencies has a dependency that breaks the production server but works on localhost. I have no idea which and it is really hard to test since I need to deploy every time (and break the server on preprod.silex.me)

src/js/service/cloud-storage.js
lexoyo: fixed

package.json
lexoyo: ok I decided to give it one more chance, I fixed all versions to the latest available and it seems to work on preprod.silex.me

:+1:

dist/server/index.js
lexoyo: :+1: 
I'll try to refactor
If you could write this in a separate issue that would be great - not a problem otherwise, I'll try

dist/server/WebsiteRouter.js
lexoyo: finally I removed it from CE2 since it doesn't need a parser (right?)
and I added it here https://github.com/silexlabs/Silex/pull/718

JbIPS: I think the documentation is outdated on this point but the changelog clearly mentioned the new way to parse bodies https://expressjs.com/en/changelog/4x.html

lexoyo:
Do you think that everything is fine now?
* package.json with fixed versions
* removed body-parser
* fixed the other things

Only the refactoring and the test of existing files remain

lexoyo:
Let's do it!

lexoyo:
Deployment in progress
Silex is down

lexoyo:
Silex is back up
This is now live, I'll have a look at the server performance statistics soon

PR #718, opened 2018-04-01
Move CE API to the route `/ce/*`, add needed body parsers  

lexoyo:


package.json
```diff
@@ -41,7 +41,6 @@
   },
   "dependencies": {
     "bluebird": "3.5.1",
-    "body-parser": "1.18.2",
```
JbIPS: Good thing you remove that but you still use it in your files 😕

lexoyo:
I think we can merge this and test on preprod.silex.me

package.json
lexoyo: Ok I found it in routers

When I change
```
const bodyParser = require('body-parser');
router.post('/tasks/:task', bodyParser.json(), (req, res, next) => {
router.put(/\/website\/(.*)\/put\/(.*)/, bodyParser.text({limit: '1mb'}), writeWebsite);
```
for 
```
app.use(express.json());
app.use(express.urlencoded());
```
And then I post JSON or put text in the body it is not parsed as text. I triple checked the content type and tried every combinations... I searched issues with urlencoded but found nothing online. There seem to be an issue though

I have no choice but put back body-parser

JbIPS:
WebsiteRouter, line 16

lexoyo:
Hello
I think we are finally good to go?

PR #791, opened 2018-08-26
WIP Build Silex native app (Electron) on travis ci

lexoyo:
it is working properly but the upload to github releases is still failing


lexoyo:
It builds for linux and osX but for windows we will need to build on appveyor
https://bretahajek.com/2016/11/electron-builder-appveyor-travis-ci-memory-king/

.travis.yml
```diff
@@ -0,0 +1,27 @@
...
+  yarn: true
+notifications:
+  email: false
+script:
+- npm run build
+- npm run build:electron
+deploy:
+  provider: releases
+  api_key:
+    secure: H72TbvbYGaf8vjRZvTZuXQMDxiOIuYskVQJQrGBF6B7F748SwH6UIBBkscXD1rVE1c55WhmuWIGGUZi1++wlhBHylwmBwfqwcbVykMiRMHMU6l9PEMIvxMOCyVZlBxxo0g3ULOJnvoakGC6NCxrJAl4jLIvRzZHMGACzTNpTei8=
```
JbIPS: Are you sure you want to put this on GitHub?

.travis.yml
```diff
@@ -0,0 +1,27 @@
...
+  on:
+    repo: lexoyo/Silex
+    branch: electron-ci
+env:
+  global:
+    - GH_TOKEN=H72TbvbYGaf8vjRZvTZuXQMDxiOIuYskVQJQrGBF6B7F748SwH6UIBBkscXD1rVE1c55WhmuWIGGUZi1++wlhBHylwmBwfqwcbVykMiRMHMU6l9PEMIvxMOCyVZlBxxo0g3ULOJnvoakGC6NCxrJAl4jLIvRzZHMGACzTNpTei8=
```
JbIPS: Put that in your secret Travis configuration and is this var in the `api_key`

.travis.yml
lexoyo: I was surprised but that's how they say we should do it... And env vars from the project settings in Travis are not set when a build is triggered by GitHub so there is no other choice

Also this is an encrypted key which can only be used by Travis, and it was generated by Travis CLI (a little modified by hand)

.travis.yml
lexoyo: I don't know what it is... And it still does not deploy. Maybe you can give it a go
It builds for Linux and Mac which is already something good

JbIPS: After reading the [manual for GitHub Releases](https://docs.travis-ci.com/user/deployment/releases/), I think you're missing 3 things:

- [ ] `skip_cleanup = true`
> Make sure you have skip_cleanup set to true, otherwise Travis CI will delete all the files created during the build, which will probably delete what you are trying to upload.
- [ ] `on.tags = true` because GH works like this and it will be a more cleaner
- [ ] `file` pointing to your executable

lexoyo:
For me it is still failing the same way
https://travis-ci.org/lexoyo/Silex/jobs/421312600

lexoyo:
This PR is obsolete with @JbIPS's work

PR #792, opened 2018-08-28
Build electron on travis CI

lexoyo:


lexoyo:
@JbIPS can I merge this?

JbIPS:
Only if you squash it 😃

lexoyo:
@JbIPS it looks like the linters actually build silex every time, the should only check the code right?

JbIPS:
Yeah I'm looking for a way to skip the installation step.

Maybe something with `install: skip`

PR #834, opened 2018-11-06
view on desktop or mobile or both

lexoyo:


lexoyo:
this is the "hide on desktop" feature!!

PR #1579, opened 2024-04-17
NPM instructions

SuperDelphi:
Updated NPM instructions (added build instruction and clarified the start instruction).


lexoyo:
Now that i read the whole paragraph I understand the previous text, it is about installing silex from npm not from source

So the old text makes more sense than the new one

PR #1580, opened 2024-04-18
Sticky top section in the Style Manager

SuperDelphi:
This contribution makes the top section of the Style Manager sticky, so that it stays visible when scrolling. It also comes with minor layout adjustments in this section, in order to welcome this new addition.

firefox_lFKTN608aZ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .idea/
 .temp
 .cache
+package-lock.json
 node_modules
 *.log
 .vscode
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.sublime-project
 *.sublime-workspace
 *.DS_Store
+.idea/
 .temp
 .cache
 node_modules
--- a/src/css/grapesjs.scss
+++ b/src/css/grapesjs.scss
@@ -92,9 +92,9 @@
 
 // Sticky CSS tags section
 
-.gjs-pn-views-container > div > div:first-child > div:first-child {
+.top-style-section {
   position: sticky;
-  top: 0;
+  top: -1px; /* Fixes a visual gap on the top */
   border-bottom: solid $lighterPrimaryColor 1px;
   background-color: $primaryColor;
   box-shadow: 0 0 15px 0 #00000060;
@@ -126,16 +126,6 @@
 
 #gjs-clm-up {
   margin-top: 0;
-}
-
-// Adapting the layout of the suggestion list
-
-.gjs-suggest {
-  max-height: 30vh;
-  overflow-y: scroll;
-  overflow-x: hidden;
-  padding-top: 7px;
-  border-bottom: solid $primaryColor 7px;
 }
 
 /*
--- a/src/css/grapesjs.scss
+++ b/src/css/grapesjs.scss
@@ -90,6 +90,44 @@
   right: var(--viewsPanelWidth);
 }
 
+// Sticky CSS tags section
+
+.gjs-pn-views-container > div > div:first-child > div:first-child {
+  position: sticky;
+  top: 0;
+  border-bottom: solid $lighterPrimaryColor 1px;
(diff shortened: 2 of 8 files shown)


SuperDelphi:
I've just added a quick fix to limit the height of the tag suggestion list.

src/css/grapesjs.scss
```diff
@@ -90,6 +90,54 @@
   right: var(--viewsPanelWidth);
 }
 
+// Sticky CSS tags section
+
+.gjs-pn-views-container > div > div:first-child > div:first-child {
```
lexoyo, line 95: This is not a good way to select the element you want
Use grapesjs API to get the element you want (or at least one of its close containers) and add a css class to it, e.g get the container with `silex.getEditor()
.Panels.getPanel('views-container')`
Please look for a better way

.gitignore
```diff
@@ -2,8 +2,10 @@
 *.sublime-project
 *.sublime-workspace
 *.DS_Store
+.idea/
```
lexoyo, line 5: why this?

.gitignore
```diff
@@ -2,8 +2,10 @@
...
 .temp
 .cache
+package-lock.json
```
lexoyo, line 8: this no
it's often removed because of the meta repo / npm workspaces but ideally we would need package-lock.json

src/css/grapesjs.scss
```diff
@@ -90,6 +90,54 @@
...
+  padding-bottom: 0;
+}
+
+#gjs-clm-up {
+  margin-top: 0;
+}
+
+// Adapting the layout of the suggestion list
+
+.gjs-suggest {
```
lexoyo, line 133: [this is a plugin, check its doc](https://github.com/silexlabs/grapesjs-ui-suggest-classes/)

You can choose to add this css in the plugin config - this is a good solution if it doesn't mean too much code and do not override the plugins' styles that we want to keep (i am not a fan of these styling options)
https://github.com/silexlabs/Silex/blob/56ccba435e553880e94f0127d56271af0c230f3c/src/ts/client/grapesjs/index.ts#L271

Or add the css in the `grapesjs-plugins.scss` file
https://github.com/silexlabs/Silex/blob/56ccba435e553880e94f0127d56271af0c230f3c/src/css/grapesjs-plugins.scss

lexoyo (changes requested):

lexoyo:
The result looks great @SuperDelphi ! 
Let's make this PR right :+1:

src/css/grapesjs.scss
SuperDelphi: Yeah, I knew it wasn't ideal, thank you for the guidance. Will do this.

.gitignore
SuperDelphi: The IntelliJ IDEA IDEs create a folder to store all of the project's metadata, but you don't want it to be added to the repository in general, for various reasons (including privacy).

.gitignore
SuperDelphi: Ok! Got it

src/css/grapesjs.scss
SuperDelphi: Thank you. I will add the CSS in the grapesjs-plugins.scss file, it seems to be the cleaner way to do it.

.gitignore
lexoyo: You should not commit this

SuperDelphi: Ok! I use to do this, but maybe that's not the ideal way to deal with this folder.

src/css/grapesjs.scss
SuperDelphi: https://github.com/silexlabs/Silex/pull/1580/commits/07dd4e03e735d5d4aa63a27fc05fb08753b91ea2
Please review this (thank you for your patience, despite my technical issues today!)

PR #1592, opened 2024-05-02
publish index-local.css in local unsecured situation

oliviermgx:


oliviermgx:
hopping the difference is easily readable ?

lexoyo (approved):

PR #1594, opened 2024-05-06
Keymaps feature

SuperDelphi:
# Description

This pull request brings several pre-defined, ready-to-use shortcuts.

# Shortcuts list

## Full panels

- Shift+Alt+S to open the Settings panel
- Shift+Alt+P to open the Publish panel
- Shift+Alt+F to open the Fonts panel
- Tab to open/close the Preview mode

## Left panels

- Shift+L to open the Layers panel
- Shift+A to open the Blocks panel ("a" for "add")
- Shift+N to open the Notifications panel
- Shift+P to open the Pages panel
- Shift+S to open the Symbols panel
- Escape to close the left panel

## Workflow-specific

- Shift+B to select the Body element


src/ts/client/grapesjs/index.ts
```diff
@@ -111,6 +113,12 @@ const QUATERNARY_COLOR = '#A291FF'
 const DARKER_PRIMARY_COLOR = '#363636'
 const LIGHTER_PRIMARY_COLOR = '#575757'
 
+// Commands
+export const cmdToggleLayers = 'open-layers'
+export const cmdToggleBlocks = 'open-blocks'
+export const cmdToggleSymbols = 'open-symbols'
+export const cmdToggleNotifications = 'open-notifications'
```
lexoyo, line 120: yes nice

src/ts/client/grapesjs/index.ts
```diff
@@ -384,8 +399,8 @@ export async function initEditor(config: EditorConfig) {
 
       // Remove blocks and layers buttons from the properties
       // This is because in Silex they are on the left
-      views.buttons.remove('open-blocks')
-      views.buttons.remove('open-layers')
+      views.buttons.remove(cmdToggleBlocks)
+      views.buttons.remove(cmdToggleLayers)
```
lexoyo, line 403: cmd* means "command" but here it's an ID your using it for

so maybe rename cmdToggleBlocks toggleBlocksId if it makes more sense?

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -0,0 +1,64 @@
+import {Editor, PluginOptions} from 'grapesjs'
+import {cmdPublish} from './PublicationUi'
+import {cmdOpenFonts} from '@silexlabs/grapesjs-fonts'
+import {cmdToggleBlocks, cmdToggleLayers, cmdToggleNotifications, cmdToggleSymbols} from './index'
+import {cmdTogglePages} from './page-panel'
+import {cmdOpenSettings} from './settings'
+
+// Utility functions
+
+const selectBody = (editor: Editor): void => {
```
lexoyo, line 10: I'd say this should be made a command and it should probably be defined in utils or somewhere where it belongs

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -0,0 +1,64 @@
...
+  const wrapper = editor.DomComponents.getWrapper()
+  editor.select(wrapper)
+}
+
+const toggleCommand = (editor: Editor, name: string): void => {
```
lexoyo, line 15: I have a question: why this toggleCommand function? It doesn't work the same if you pass a command to km.add ?

I ask because toggleCommand takes a command and implements something similar to grapes' [stateful commands](https://grapesjs.com/docs/modules/Commands.html#stateful-commands) right?

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -0,0 +1,64 @@
...
+  km.add('general:open-settings', 'shift+alt+s', editor => toggleCommand(editor, cmdOpenSettings))
+  km.add('general:open-publish', 'shift+alt+p', editor => toggleCommand(editor, cmdPublish))
+  km.add('general:open-fonts', 'shift+alt+f', editor => toggleCommand(editor, cmdOpenFonts))
+  km.add('panel:preview', 'space', editor => toggleCommand(editor, 'preview'))
+  km.add('panel:layers', 'shift+l', editor => toggleCommand(editor, cmdToggleLayers))
+  km.add('panel:blocks', 'shift+a', editor => toggleCommand(editor, cmdToggleBlocks))
+  km.add('panel:notifications', 'shift+n', editor => toggleCommand(editor, cmdToggleNotifications))
+  km.add('panel:pages', 'shift+p', editor => toggleCommand(editor, cmdTogglePages))
+  km.add('panel:symbols', 'shift+s', editor => toggleCommand(editor, cmdToggleSymbols))
+  km.add('panel:close', 'escape', resetPanel)
```
lexoyo, line 59: i think it would be better to use tab

and keep space for moving around the canvas like in adobe's software

src/ts/client/grapesjs/index.ts
SuperDelphi: Will do that! Makes sense.

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: I agree!

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: I had tried using only the command. It doesn't work as expected, as it will open panels without closing the other ones, mainly. You have to stop the command associated with a panel to close it.

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: Yes, it makes sense! I was inspired by Blender's way to do it, I believe.

src/ts/client/grapesjs/index.ts
SuperDelphi: Oh by the way, the reason I named them like that is because I've seen this convention at some other places in the code. I was in the process of maintaining the coherence.

Examples:
image
image
Etc.

Let me know what is your verdict.

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: All things considered, I don't think it's a good idea to override Tab, since it is used for accessibility purposes to move focus between elements. What do you think?

src/ts/client/grapesjs/index.ts
lexoyo: ok let's keep it coherent then :+1:

src/ts/client/grapesjs/keymaps.ts
lexoyo: The space bar will definitely be needed at one point to move around

I don't think it makes sense to use silex with accessibility features but maybe i'm wrong

So maybe ctrl+shift+P or something

lexoyo (changes requested):

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: It has been! When I update the branch, a command declaration will be written in ``keymaps.ts`` and the selectBody() function will be in ``utils.ts``

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: Ok! I think I tried to define "tab" but I couldn't trigger it, will retry. If not, I will add it on Shift+Alt+P, maybe (to stay coherent with the other keymaps, and also because Ctrl+Shift+P triggers the opening of a private browsing window in Firefox).

SuperDelphi: Nevermind, Tab works well!

SuperDelphi:
The branch has been updated!

lexoyo (changes requested):
these things are still missing I believe for the feature to be complete

* [ ] while in the text editor, `esc` should leave the text edition
* [ ] the titles on the buttons should inform the user that there is a shortcut - cf screenshot bellow

my feedback about the choice of shortcuts, to be discussed

* [ ] you where right, tab is a bad idea for preview, we do use tab for tab, sorry my bad
* [ ] I would avoid shortcuts with 2 modifiers when possible? e.g "Full panels" could be alt+ instead of shift+alt+?
* [ ] `esc` should close the last opened dialog - e.g. open publish, then esc, then close publish. Also right now it closes all pannels, e.g. i open the pages panel, then settings dialog, then esc => both are closed which is not what i intended (only settings should close)
* [ ] `esc` in the settings when a text field is focused should close the settings, but instead it does nothing
* [ ] good addition would be to pay attention to focus when a dialog opens (it is out of this feature I think, it should work also when open with a button), e.g. publish dialog focus the publish button

also where should i suggest new shortcuts? here? should we do a special discussion in the forums? for now i have

* responsive switch (desktop, tablet, ...)
* a shortcut to add the most used elements (text, container, section) => add it to the selected element
* check https://github.com/silexlabs/Silex/wiki/Keyboard-Shortcuts

SuperDelphi:
@lexoyo I added some functionality to my PR!

✅ There are shortcuts hints in the project bar buttons (including the Publish button).
✅ The ``Shift+Alt+...`` have been replaced with ``Alt+...``.
✅ Publish/Apply buttons are now automatically focused when the modal or the Publish dialog are opened, which can be triggered with ``Enter``.
ℹ️ The editor's input fields can be exited with the ``Escape`` key, **except for (Rich) Text components** (this is a complex issue).
ℹ️ The ``Escape`` button is "smart" now. It will only close the foremost context. However, when exiting a "full panel" (ex: the Settings panel), even if you opened a "left panel" beforehand, the latter will be closed. It's an already existing issue due to the fact "left panels" close when a "full panel" opens. It is its own issue (that we can solve in another PR).

src/ts/client/grapesjs/index.ts
```diff
@@ -289,6 +305,13 @@ export function getEditorConfig(config: ClientConfig): EditorConfig {
       [filterStyles]: {
         appendBefore: '.gjs-sm-sectors',
       },
+      [internalLinksPlugin.toString()]: {
```
lexoyo, line 308: i don't understand, why did you have to add this? it wasn't already there?

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -0,0 +1,111 @@
...
+  kmBlocks: prefixKey + '+A',
+  kmNotifications: prefixKey + '+N',
+  kmPages: prefixKey + '+P',
+  kmSymbols: prefixKey + '+S',
+  kmClosePanel: 'Escape'
+}
+
+// Utility functions
+
+const toggleCommand = (editor: Editor, name: string): void => {
```
lexoyo, line 27: it is not very important but you probably want to use function ABCD() {} insead of const ABCD = () => {}

because arrow function makes no sense here without a context

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -0,0 +1,111 @@
...
+  // Commands
+  editor.Commands.add(cmdSelectBody, selectBody)
+
+  if (opts.disableKeymaps) return
+  if (opts.prefixKey) prefixKey = opts.prefixKey
+
+  const km = editor.Keymaps
+
+  // Panels
+  km.add('general:open-settings', defaultKms.kmOpenSettings.toLowerCase(), editor => toggleCommand(editor, cmdOpenSettings))
```
lexoyo, line 80: it would be easier to understand if defaultKms is an array of objects with [keymap::add params](https://grapesjs.com/docs/api/keymaps.html#add) id, keys, handler and opts attributes

then you could loop over tye array and save the code bellow?

lexoyo: or maybe keep defaultKms an object, with the IDs as keys and keys, handler and opts as attributes

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -0,0 +1,111 @@
...
+  km.add('panels:close-panel', defaultKms.kmClosePanel.toLowerCase(), escapeContext)
+
+  // Handling the Escape keymap during text edition
+  document.addEventListener('keydown', event => {
+    if (event.key === defaultKms.kmClosePanel) {
+      const target = event.target as HTMLElement | null
+      const rte = editor.getEditing()
+
+      if (rte) { // If in Rich Text edition...
+        // TODO: Close the Rich Text edition and un-focus the text field
```
lexoyo, line 98: this is still to do right?

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -0,0 +1,111 @@
...
+          escapeContext(editor)
+        } else if (isTextOrInputField(target)) { // If it's a text field...
+          target.blur()
+        }
+      }
+    }
+  })
+
+  // Workflow-specific keymaps
+  km.add('workflow:select-body', prefixKey + '+b', cmdSelectBody)
```
lexoyo, line 110: why not make it `esc` to select the body? maybe only when the dialog stack is empty

src/ts/client/grapesjs/settings.ts
```diff
@@ -59,7 +59,6 @@ export const settingsDialog = (editor, opts) => {
         editor.trigger(ClientEvent.SETTINGS_SAVE_END, page)
         editor.stopCommand(cmdOpenSettings)
       }
-      form.querySelector('input')?.focus()
```
lexoyo, line 62: actually this makes sense? will it focus the first input ?

src/ts/client/utils.ts
```diff
@@ -45,5 +45,19 @@ export async function hashString(str: string): Promise<string> {
...
+ * Select the <body> element in the editor.
+ * @param editor The GrapesJS editor.
+ */
+
+export function selectBody(editor: Editor): void {
+  editor.select(editor.DomComponents.getWrapper())
+}
+
+export function isTextOrInputField(element: HTMLElement): boolean {
+  return element.tagName === 'TEXTAREA' ||
```
lexoyo, line 61: you probably want to include the select? maybe other form elements?

lexoyo (changes requested):

src/ts/client/grapesjs/index.ts
SuperDelphi: It is the result of a merge, it was indeed already there 👍🏼

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: I thought about this, but if you look at the content of ``defaultKms``, you will see that I put majuscules. It is because I use them for keyboard hints, but you need to lower the case before declaring the keymaps.
What I could do is make it an object with all the properties needed, and when I need a "cleaner" display, I can add a conversion function into ``utils.ts``.

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: Yes! (Please refer to my 4th ✅)

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: I think Jonas wanted a shortcut for this when he's editing HTML attributes, he's doing a back-and-forth. I can add this, but it seems a bit unintuitive (in my opinion) and not consistent, since it won't work when we have a left panel (which is often the case, I imagine?), so it will add friction. Tell me what's your verdict!

src/ts/client/grapesjs/settings.ts
SuperDelphi: I removed it because it was in conflict with the automatic focus of the primary button of the modal, but maybe you only wanted this for dialogs? If so, I'll remove this behaviour and rollback this line.

src/ts/client/utils.ts
SuperDelphi: Yes, good idea!

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: By the way, I see you also use arrow functions "without context" for main plugin functions (ex: project-bar.ts), do I change all of the arrow functions into "classical" functions?

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: (Done)

src/ts/client/grapesjs/keymaps.ts
lexoyo: The back and forth is a workaround because there is a bug in some expression inputs, so don't take it into account, I will fix the bug instead

I think esc for selecting the body is good but only when we are not in a rich text 

Maybe we can try and test it?

src/ts/client/grapesjs/settings.ts
lexoyo: yes this is better to focus the first input, as it is in a form. So enter will submit the form, esc cancel, tab go to next input

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: I can try, sure. I'm considering we that we also keep ``Shift+B`` (?)

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -0,0 +1,155 @@
...
+import {cmdOpenFonts} from '@silexlabs/grapesjs-fonts'
+import {cmdToggleBlocks, cmdToggleLayers, cmdToggleNotifications, cmdToggleSymbols} from './index'
+import {cmdTogglePages} from './page-panel'
+import {cmdOpenSettings} from './settings'
+import {isTextOrInputField, selectBody} from '../utils'
+import {PublishableEditor} from './PublicationManager'
+
+// Utility functions
+
+function getPanelCommandIds (): string[] {
```
SuperDelphi, line 12: I can't directly declare an array because there is a circular dependency between "index.ts" and "keymaps.ts".

lexoyo (approved):

PR #1596, opened 2024-05-20
WIP : pour analyse suite modif gitlabConnector par #1589

oliviermgx:
WIP : pour analyse suite modif gitlabConnector par #1589

### DIFF
--- a/src/ts/client/grapesjs/PublicationManager.ts
+++ b/src/ts/client/grapesjs/PublicationManager.ts
@@ -298,6 +298,11 @@ export class PublicationManager {
         data: data as ApiPublicationPublishBody,
         options: this.settings.options,
       })
+      
+      // in gitlab pages situation, getUrl from gitlabHostingConnector gives publication url obtained with Gitlab API pages
+      console.info('Gitlab url: ', url)
+      // could be used in an future UI
+
       this.job = job
       this.status = jobStatusToPublicationStatus(this.job.status)
       this.trackProgress()
--- a/src/ts/plugins/server/GitlabHostingConnector.ts
+++ b/src/ts/plugins/server/GitlabHostingConnector.ts
@@ -22,19 +22,15 @@
  * @see https://docs.gitlab.com/ee/user/project/pages/getting_started/pages_ci_cd_template.html
  */
 
-import GitlabConnector, { GitlabOptions, GitlabToken } from './GitlabConnector'
+import GitlabConnector, { GitlabOptions, GitlabSession} from './GitlabConnector'
 import { HostingConnector, ConnectorFile } from '../../server/connectors/connectors'
 import { ConnectorType, WebsiteId, JobData, JobStatus, PublicationJobData } from '../../types'
 import { JobManager } from '../../server/jobs'
 import { join } from 'path'
 import { ServerConfig } from '../../server/config'
 
-//type GitlabSession = ConnectorSession
-//type GitlabOption = GitlabOptions
-type GitlabSession = Record<string, GitlabToken>
+export default class GitlabHostingConnector extends GitlabConnector implements HostingConnector {
 
-export default class GitlabHostingConnector extends GitlabConnector implements HostingConnector/*<GitlabSession>*/ {
-  connectorId = 'gitlab_hosting'
   displayName = 'Gitlab hosting'
   connectorType = ConnectorType.HOSTING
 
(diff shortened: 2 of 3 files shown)


src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -0,0 +1,109 @@
...
+        jobError(job.jobId, message)
+      }
+    })
+    return job
+  }
+
+  async getUrl(session: GitlabSession, websiteId: WebsiteId): Promise<string> {
+    const id = websiteId
+    const filePath = join(this.options.domain, id, 'index.html')
+    const fileUrl = new URL(filePath, 'file://')
```
lexoyo, line 106: This doesn't make sense for the gitlab connector
Please check in silex code base where this getUrl method is called? Then we see what to return

src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -0,0 +1,109 @@
...
+import GitlabConnector, { GitlabOptions, GitlabToken } from './GitlabConnector'
+import { HostingConnector, ConnectorFile } from '../../server/connectors/connectors'
+import { ConnectorType, WebsiteId, JobData, JobStatus, PublicationJobData } from '../../types'
+import { JobManager } from '../../server/jobs'
+import { join } from 'path'
+import { ServerConfig } from '../../server/config'
+
+//type GitlabSession = ConnectorSession
+//type GitlabOption = GitlabOptions
+type GitlabSession = Record<string, GitlabToken>
```
lexoyo, line 34: this is already in GitlabConnector.ts, you probably want to export it from there and import it here

src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export default class GitlabHostingConnector extends GitlabConnector implements HostingConnector/*<GitlabSession>*/ {
+  connectorId = 'gitlab_hosting'
```
lexoyo, line 37: the session problems you have probably come from here

i see [here in GitlabConnector::getSessionToken](https://github.com/silexlabs/Silex/blob/dev/src/ts/plugins/server/GitlabConnector.ts#L397) that it uses `connectorId`

So your problem comes from there: the session is in `{ gitlab: { token } }` but in your instance it will look for `{ gitlab_hosting: { token } }` 

I believe you can should remove this line and it will work as expected (it will use `gitlab` as connectorId)

lexoyo (changes requested):

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: done, ok thx

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: yes, working, thx

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: I don't understand, it seems to be used in connector.ts and DownloadConnector.ts ?

lexoyo: this is where it is used
https://github.com/silexlabs/Silex/blob/549bf7b6b2419113718b98b6ab97d0280252e871/src/ts/server/api/publicationApi.ts#L140

so after publication it will send an url to the client side, which gets it here
https://github.com/silexlabs/Silex/blob/549bf7b6b2419113718b98b6ab97d0280252e871/src/ts/client/grapesjs/PublicationManager.ts#L294

but then it is never used anywhere :face_with_diagonal_mouth: 

we could use this URL to display a big button "Go to your website" but this will be another feature for later

so for now, you should return the URL of the website. Maybe move the code which calls gitlab api there?

oliviermgx: Yes, clearer, I'll do that

src/ts/client/grapesjs/PublicationManager.ts
```diff
@@ -298,6 +298,11 @@ export class PublicationManager {
         data: data as ApiPublicationPublishBody,
         options: this.settings.options,
       })
+      
+      // in gitlab pages situation, getUrl from gitlabHostingConnector gives publication url obtained with Gitlab API pages
+      console.info('Gitlab url: ', url)
+      // could be used in an future UI
+
```
SuperDelphi, line 305: You might want to delete these lines after testing?

oliviermgx: It's a comment for next version
up to now, it gives url in console
could be replaced soon with à button or other UI solution as suggested by Alex above
your opinion ?

SuperDelphi: Ok understood! I think later we could add methods to allow connectors to display buttons and other elements in the publication UI after the job.

PR #1599, opened 2024-05-27
New shortcuts

SuperDelphi:
This pull request brings the following shortcuts:
- ``R`` to open the Style Manager ;
- ``T`` to open the Traits Manager ;
- ``Ctrl+D`` to duplicate one or multiple components ;
- ``Ctrl+Alt+P`` to publish the website.

It also fixes a bug in the Publish dialog where it would cause an error when closed if it hadn't been opened with the usual button.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@silexlabs/silex",
   "description": "Free and easy website builder for everyone.",
-  "version": "3.0.0-alpha.159",
+  "version": "3.0.0-alpha.158",
   "version:backwardcompat": "3.0.0",
   "author": {
     "name": "Alex Hoyau",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@silexlabs/silex",
   "description": "Free and easy website builder for everyone.",
-  "version": "3.0.0-alpha.158",
+  "version": "3.0.0-alpha.159",
   "version:backwardcompat": "3.0.0",
   "author": {
     "name": "Alex Hoyau",
--- a/src/ts/client/grapesjs/PublicationUi.ts
+++ b/src/ts/client/grapesjs/PublicationUi.ts
@@ -327,7 +327,9 @@ export class PublicationUi {
   async closeDialog() {
     this.isOpen = false
     this.renderDialog(null, null)
-    if(this.sender) this.sender.set('active', 0) // Deactivate the button to make it ready to be clicked again
+    if (this.sender && typeof this.sender.set === 'function') {
+      this.sender.set('active', 0)
+    } // Deactivate the button to make it ready to be clicked again
     this.sender = null
   }
   async toggleDialog() {
--- a/src/ts/client/grapesjs/keymaps.ts
+++ b/src/ts/client/grapesjs/keymaps.ts
@@ -144,8 +144,8 @@ export function keymapsPlugin(editor: Editor, opts: PluginOptions): void {
   // Default keymaps
   for (const keymap in defaultKms) {
     km.add(defaultKms[keymap].id, defaultKms[keymap].keys, defaultKms[keymap].handler, {
-      ...defaultKms[keymap].options,
-      prevent: true
+      prevent: true,
+      ...defaultKms[keymap].options
     })
   }
 
--- a/src/ts/client/grapesjs/keymaps.ts
+++ b/src/ts/client/grapesjs/keymaps.ts
@@ -5,7 +5,6 @@ import {cmdTogglePages} from './page-panel'
 import {cmdOpenSettings} from './settings'
 import {isTextOrInputField, selectBody} from '../utils'
 import {PublishableEditor} from './PublicationManager'
-import {cmdPublish} from './PublicationUi'
 
 function getPanelCommandIds (): string[] {
   return [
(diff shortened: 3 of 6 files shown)


src/ts/plugins/server/GitlabHostingConnector.ts
SuperDelphi, line 1: I just don't understand why my PR also carries these changes.

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -69,7 +112,7 @@ export const defaultKms = {
   kmOpenPublish: {
     id: 'general:open-publish',
     keys: 'alt+p',
-    handler: editor => toggleCommand(editor, cmdPublish)
+    handler: openPublish
```
SuperDelphi, line 115: The ``publish`` command doesn't change the button's state, hence why I don't use the command directly.

src/ts/client/grapesjs/PublicationUi.ts
```diff
@@ -327,7 +327,9 @@ export class PublicationUi {
   async closeDialog() {
     this.isOpen = false
     this.renderDialog(null, null)
-    if(this.sender) this.sender.set('active', 0) // Deactivate the button to make it ready to be clicked again
+    if (this.sender && typeof this.sender.set === 'function') {
+      this.sender.set('active', 0)
+    } // Deactivate the button to make it ready to be clicked again
```
SuperDelphi, line 332: If the sender doesn't have a ``set()`` function (for example, if it's not a button), we don't run this instruction.

SuperDelphi (commented):
I've left comments to justify some choices!

lexoyo:
> I've left comments to justify some choices!

I guess the best way to do this is to comment using github on your code.

SuperDelphi:
Yes, I'm talking about GitHub comments, sorry.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@silexlabs/silex",
   "description": "Free and easy website builder for everyone.",
-  "version": "3.0.0-alpha.158",
+  "version": "3.0.0-alpha.159",
```
lexoyo, line 4: Don't increment it by hand, it will be done when a new release is started

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -32,6 +37,43 @@ function toggleCommand (editor: Editor, name: string): void {
   }
 }
 
+/**
+ * Opens the Traits Manager panel.
+ * @param editor The editor.
+ */
+function openTraitsManager(editor: Editor): void {
+  editor.Panels.getButton('views', 'open-tm').set('active', true)
```
lexoyo, line 45: In grapesjs code I found this, doesn't it mean that there is a command for that?
https://github.com/GrapesJS/grapesjs/blob/b247a580751b626017c44ddac2981b9dff6772df/src/commands/index.ts#L60

I'd rather avoid programatically click on the button

src/ts/client/grapesjs/keymaps.ts
lexoyo: Could you fix the publish command instead ?
Or leave it as is but create an issue

It's better than a workaround in the code

src/ts/plugins/server/GitlabHostingConnector.ts
lexoyo: I don't see any change here 🤷

lexoyo (changes requested):

package.json
SuperDelphi: I think it automatically incremented, I haven't touched it (maybe because I did ``npm run build`` to run tests). I'm changing this.

src/ts/client/grapesjs/keymaps.ts
SuperDelphi: This is indeed the command to open the Traits Manager (there's another one for the Style Manager). This is the first thing I tried to use, but it will only open the panel, **and not update the button states**.

If I understand correctly, it's not really like a ``button.click()``, but more like enabling the command bound to the button. What should we do?

lexoyo:
Also if it is simple could you update the title on the "open traits" and "open properties" buttons with the shortcut
Screenshot from 2024-05-27 17-05-42

src/ts/client/grapesjs/keymaps.ts
```diff
@@ -64,48 +56,57 @@ export const defaultKms = {
...
+  },
+  kmStyleManager: {
+    id: 'panels:style-manager',
+    keys: 'r',
+    handler: editor => setButton(editor, 'views', 'open-sm', true)
+  },
+  kmTraitsManager: {
+    id: 'panels:traits',
+    keys: 't',
+    handler: editor => setButton(editor, 'views', 'open-tm', true)
```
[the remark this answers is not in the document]
lexoyo: these could be a comment in the code ;)

package.json
lexoyo: please keep it to the original
at least don't commit it

lexoyo:
> Also if it is simple could you update the title on the "open traits" and "open properties" buttons with the shortcut Screenshot from 2024-05-27 17-05-42

what about this comment?

lexoyo (changes requested):

lexoyo (approved):

PR #1604, opened 2024-06-07
Bug job id return in gitlab host connector

oliviermgx:


src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -180,9 +200,10 @@ export default class GitlabHostingConnector extends GitlabConnector implements H
     } catch (error) {
       console.error('Error during creating new tag:', error.message)
       jobError(job.jobId, `Failed to create new tag: ${error.message}`)
-      return false
+      return 'Failed'
```
lexoyo, line 203: I suggest to return null in case there is no tag
So the function will have a return type of string | null

src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -146,12 +146,32 @@ export default class GitlabHostingConnector extends GitlabConnector implements H
...
-  async getGitlabJobLogsUrl(session: GitlabSession, websiteId: WebsiteId, projectUrl: string): Promise<string> {
-    const jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
-    return `${projectUrl}/-/jobs/${jobs[0].id}`
+  async getGitlabJobLogsUrl(session: GitlabSession, websiteId: WebsiteId, job: PublicationJobData, { startJob, jobSuccess, jobError }: JobManager, projectUrl: string, tag): Promise<string> {
+    let jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
+    // waiting for the job corresponding to the current tag
+    let i = 0
+    setTimeout (() => {
+      while (jobs[0].ref !== tag && i<20) {
+        jobs = this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
```
lexoyo, line 155: Although there's a condition to limit the loop to 20 iterations, the loop does not include any delay or await between iterations. It will make many calls to callApi for each timeout

src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -146,12 +146,32 @@ export default class GitlabHostingConnector extends GitlabConnector implements H
     return `${projectUrl}/pages`
   }
 
-  async getGitlabJobLogsUrl(session: GitlabSession, websiteId: WebsiteId, projectUrl: string): Promise<string> {
-    const jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
-    return `${projectUrl}/-/jobs/${jobs[0].id}`
+  async getGitlabJobLogsUrl(session: GitlabSession, websiteId: WebsiteId, job: PublicationJobData, { startJob, jobSuccess, jobError }: JobManager, projectUrl: string, tag): Promise<string> {
+    let jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
```
lexoyo, line 150: Don't make a first call then many other ones

I suggest :

1. Do
1. await get the tag (callApi)
1. await TIME_BEFORE_RETRY - see [the way to do this (first solution of this article)](https://attacomsian.com/blog/javascript-sleep-delay-wait-pause)
1. While elapsed time < TIMEOUT and the tag is not the expected one

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: I don't think so, the setTimeout is tune with a delay and I have mesured the mean number f call that are between 0 and 5 cycles. What I am missing ?

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: and tested at the beginning with `if (!successTag)` instead of  `if (successTag === 'Failed')` ?

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: "Don't make a first call then many other ones"
I don't understand, I need to get the first time the jobs in order to read its jobs.ref containing the current tag ? otherwise I don't know how to initialize the while loop ?

oliviermgx: "await TIME_BEFORE_RETRY ", I thought that setTimeout do the same ? https://www.freecodecamp.org/news/javascript-settimeout-how-to-set-a-timer-in-javascript-or-sleep-for-n-seconds/

oliviermgx: "await get the tag (callApi)
await TIME_BEFORE_RETRY - see [the way to do this (first solution of this article)](https://attacomsian.com/blog/javascript-sleep-delay-wait-pause)
While elapsed time < TIMEOUT and the tag is not the expected one"

I beleived I'm doing exactly this algo ?

lexoyo: >> "Don't make a first call then many other ones"

> I don't understand, I need to get the first time the jobs in order to read its jobs.ref containing the current tag ? otherwise I don't know how to initialize the while loop ?

use a do while instead of a while

lexoyo: > "await TIME_BEFORE_RETRY ", I thought that setTimeout do the same ? https://www.freecodecamp.org/news/javascript-settimeout-how-to-set-a-timer-in-javascript-or-sleep-for-n-seconds/

the setTimeout takes a callback, and given the complexity of this algorithm you should use await and no callback, it will make it obvious my other feedback

src/ts/plugins/server/GitlabHostingConnector.ts
lexoyo: you don't wait for an answer (await)
so jobs will be a promise, not the actual result
and it will run 100 times without waiting for a result

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: thank you for your time to explain to me

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: just to verify my comprehension, this should work better ? 
`    setTimeout (async () => {
      while (jobs[0].ref !== tag && i<20) {
        jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
        i++
      }
    }, 100)`

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: I will change with do while, thanks for advices

src/ts/plugins/server/GitlabHostingConnector.ts
lexoyo: no because the setTimeout will not wait that the async finishes, you have multiple calls in parallel

check my proposal of algorithm, this will do a call, wait for the response, wait a little, loop

lexoyo: > check my proposal of algorithm, this will do a call, wait for the response, wait a little, loop

your algo will do: wait a little, start a call but don't get the result and don't wait, loop

oliviermgx: sorry for my lack of skills in async js field. I will try again !

src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -146,12 +147,25 @@ export default class GitlabHostingConnector extends GitlabConnector implements H
...
-    return `${projectUrl}/-/jobs/${jobs[0].id}`
+  // waiting for the job corresponding to the current tag
+  async getGitlabJobLogsUrl(session: GitlabSession, websiteId: WebsiteId, job: PublicationJobData, { startJob, jobSuccess, jobError }: JobManager, projectUrl: string, tag): Promise<string> {
+    let i = 0
+    do {
+      i++
+      const jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
+      if (jobs[0].ref === tag) {return `${projectUrl}/-/jobs/${jobs[0].id}`}
+      await setTimeout(5)
+    } while (i<19)
```
lexoyo, line 158: i think it would be clearer to compare execution time instead of the number of loops

use Date.now() to get the current time

src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -146,12 +147,25 @@ export default class GitlabHostingConnector extends GitlabConnector implements H
...
-    const jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
-    return `${projectUrl}/-/jobs/${jobs[0].id}`
+  // waiting for the job corresponding to the current tag
+  async getGitlabJobLogsUrl(session: GitlabSession, websiteId: WebsiteId, job: PublicationJobData, { startJob, jobSuccess, jobError }: JobManager, projectUrl: string, tag): Promise<string> {
+    let i = 0
+    do {
+      i++
+      const jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
+      if (jobs[0].ref === tag) {return `${projectUrl}/-/jobs/${jobs[0].id}`}
+      await setTimeout(5)
```
lexoyo, line 157: `5` here is in milliseconds, maybe you should give it more time?

lexoyo (changes requested):

src/ts/plugins/server/GitlabHostingConnector.ts
oliviermgx: I could but I have tested and it works between 1 and less than 10 * 5 ms ? I could put 50 ms if it's better ? else ?

src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -146,12 +147,23 @@ export default class GitlabHostingConnector extends GitlabConnector implements H
...
-    const jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
-    return `${projectUrl}/-/jobs/${jobs[0].id}`
+  // waiting for the job corresponding to the current tag
+  async getGitlabJobLogsUrl(session: GitlabSession, websiteId: WebsiteId, job: PublicationJobData, { startJob, jobSuccess, jobError }: JobManager, projectUrl: string, tag): Promise<string> {
+    const t0 = Date.now()
+    do {
+      const jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
+      if (jobs[0].ref === tag) {return `${projectUrl}/-/jobs/${jobs[0].id}`}
+      await setTimeout(5000)
+    } while ((Date.now() - t0) < 15000)
```
lexoyo, line 157: laaaast thing, you may want to set this timeout as an option?

lexoyo (changes requested):

lexoyo (approved):
Great! Well done!!

PR #1605, opened 2024-06-18
GitLab policy error fix

SuperDelphi:
This pull request allows the GitLab hosting connector to handle a case where the publication fails because the GitLab user is recent. The connector will check if it's plugged to the official GitLab instance, and if so, it will display a clear message inside the publication dialog, with a clickable link.

On a side note, the loading bar (<progress> element) inside the publication dialog has been slightly improved.

demo


package.json
```diff
@@ -108,9 +108,9 @@
     "@silexlabs/grapesjs-loading": "1.0.9",
     "@silexlabs/grapesjs-notifications": "^0.0.8",
     "@silexlabs/grapesjs-storage-rate-limit": "^1.0.8",
-    "@silexlabs/grapesjs-symbols": "1.0.41",
+    "@silexlabs/grapesjs-symbols": "1.0.40",
```
lexoyo, line 111: oups

you probably don't want to commit package.json

src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -146,8 +163,9 @@ export default class GitlabHostingConnector extends GitlabConnector implements H
     return `${projectUrl}/pages`
   }
 
-  async getGitlabJobLogsUrl(session: GitlabSession, websiteId: WebsiteId, projectUrl: string): Promise<string> {
+  async getGitlabJobLogsUrl(session: GitlabSession, websiteId: WebsiteId, projectUrl: string): Promise<string | undefined> {
     const jobs = await this.callApi(session, `api/v4/projects/${websiteId}/jobs`, 'GET')
+    if (!jobs.length) return undefined
```
lexoyo, line 168: Please use `null` instead of undefined, which means this is intentional

lexoyo (changes requested):
Minor changes, but olivier was working on the same piece of code, you will have to rebase on main (i already have merged his work)

Thank you for this important contribution (both bug fix and ui) !

SuperDelphi (commented):
I've merged the changes of Olivier and addressed the problems you pointed out!

lexoyo (approved):

PR #1606, opened 2024-06-19
with tag name format controlled

oliviermgx:
reg exp /\bv[0-9]+\.[0-9]+\.[0-9]+\b/ in order to verify tag name incrementable

### DIFF
--- a/src/ts/plugins/server/GitlabHostingConnector.ts
+++ b/src/ts/plugins/server/GitlabHostingConnector.ts
@@ -189,8 +189,14 @@ export default class GitlabHostingConnector extends GitlabConnector implements H
       // Fetch the latest tag
       tags = await this.callApi(session, `api/v4/projects/${projectId}/repository/tags`, 'GET')
       const latestTag = tags[0]?.name || 'v0.0.0'
-      const [major, minor, patch] = latestTag.slice(1).split('.').map(Number)
-      newTag = `v${major}.${minor}.${patch + 1}`
+      const validRex = /\bv[0-9]+\.[0-9]+\.[0-9]+\b/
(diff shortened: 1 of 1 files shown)


src/ts/plugins/server/GitlabHostingConnector.ts
```diff
@@ -189,8 +189,14 @@ export default class GitlabHostingConnector extends GitlabConnector implements H
...
       const latestTag = tags[0]?.name || 'v0.0.0'
-      const [major, minor, patch] = latestTag.slice(1).split('.').map(Number)
-      newTag = `v${major}.${minor}.${patch + 1}`
+      const validRex = /\bv[0-9]+\.[0-9]+\.[0-9]+\b/
+      if (latestTag.search(validRex) === 0) {
+        const [major, minor, patch] = latestTag.slice(1).split('.').map(Number)
+        newTag = `v${major}.${minor}.${patch + 1}`
+      } else {
+        console.error('Invalid tag name, unable to increment patch, should be : vx.y.z, like v1.2.0 \n===> Please modify it in your gitlab tag page')
+        throw new Error('Invalid tag name, unable to increment patch, should be : vx.y.z, like v1.2.0 \n===> Please modify it in your gitlab tag page')
```
lexoyo, line 198: the problem with that is that we don't allow the user to create arbitrary tags

i'm not sure what could be the best solution

1. fail like you did but add a link to the tags in gitlab and explain what to do (maybe create a tag with proper version)
2. in all the tags, take the first that has a proper format and use this one (or create a v0 if there is none)

I vote for 2 because it will never fail, hoping it will work as expected

PR #1612, opened 2024-07-23
Adding 'grapesjs-navbar' plugin

JeremyZXi:
adding grapesjs-navbar


lexoyo:
Thank you @JeremyZXi for taking the time to add it to Silex 🙏

PR #1681, opened 2026-03-29
Expose background-image property in Decorations panel #1666

### LINKED ISSUE #1666: Docs template: getting-started list uses bg-image instead of bg-color
In [the "Docs First" template](https://www.silex.me/templates/docs-first/), the 'Getting Started' list item uses a `background-image` property where a simple `background-color` would be more appropriate and performant, also this makes the behavior bugguy

To reproduce: [fork the "dev docs" templates from silex site](https://www.silex.me/templates/), then click on a container in the list "Getting started", in the element styles, open the section "Decorations" and you will see the background color is not visible but an image is set (which is invisible)

The template opened in silex, with the decorations section opened

This is a work to do in a template, you'll need to [open a template (edit in silex button)](https://www.silex.me/templates/) then modify it and then click "Contribute changes" in the site settings:

Silex site settings, with a Contribute changes link

cc @ceubri

SyedIshmumAhnaf:
## Description
This PR addresses **Issue #1666**. It registers the `background-image` property within the `silex-lib` Style Manager under the **Decorations** sector.

### The Problem
Previously, the `background-image` property was not visible in the editor UI. If a template (like "Docs First") had a residual or accidental background image set in its JSON data, it would overshadow the `background-color` property. Because the field was hidden, users had no way to see, debug, or clear this property, leading to "invisible" style bugs.

### The Fix
- Modified `packages/silex-lib/src/ts/client/grapesjs/css-props.ts`.
- Registered `background-image` as a `text` property.
- Positioned it at index 1 (directly below `background-color`) for better UX.
- Added an info hint explaining how to clear the property (by setting it to `none` or `initial`).

## Related Issues
Fixes #1666

## Testing
- Rebuilt `silex-lib` and confirmed the field appears correctly in the Decorations panel.


lexoyo:
Hello @SyedIshmumAhnaf 

> Modified packages/silex-lib/src/ts/client/grapesjs/css-props.ts

You probably want to do a PR in `https://github.com/silexlabs/silex-lib` then, but this PR is on the meta repo `https://github.com/silexlabs/Silex`

SyedIshmumAhnaf:
Thank you for the feedback, Alex! That makes perfect sense. I will close this PR and move the fix over to the silex-lib repository right away.

PR #1733, opened 2026-06-18
fix: exit with error code if package build fails

MatrixNeoKozak:
The `build-packages.js` script currently continues building subsequent packages if one fails, which can lead to cascading errors because downstream packages depend on successfully built upstream packages. This change ensures that if a package's build script fails, the entire process terminates immediately with a non-zero exit code. This prevents broken builds from being reported as successful and simplifies troubleshooting by stopping the build process as soon as a failure is encountered.


lexoyo:
I agree, it's better like that
Im working on a pretty big refactoring which will make it easier to contribute (I'll keep your changes of course)
I plan to have only one repo with everything (from meta-repo to mono-repo)

PR #1778, opened 2026-07-13
fix(desktop): ship desktop from v* releases so the updater feed stays valid

lexoyo:
Desktop ships from `v*` tags (the `desktop-v*` channel was unused) so `releases/latest` always has a valid `latest.json`; also publishes version-less installer aliases for permanent download links.


lexoyo:

PR #1779, opened 2026-07-13
fix(desktop): correct Linux package metadata (license, category, icons)

lexoyo:
Fix Linux package metadata: AGPL license, Development menu category, standard icon sizes.


lexoyo:

PR #1780, opened 2026-07-13
fix(ci): canary on prerelease tags, prod on stable tags

lexoyo:
v*-canary* tags => canary.silex.me
v* tags => v3.silex.me + latest for silex desktop downloads


lexoyo:

PR #1802, opened 2026-07-25
Fix/supabase keep edge types

lexoyo:
Fix the supabase backend type, check this discussion: https://community.silex.me/d/302-i-cant-use-dynamic-data-from-supabase-in-silex/7


lexoyo:
Probably these fields could be cleaned up?

image

image

image

PR #1807, opened 2026-07-26
 Add horizontal center button to component toolbar

### LINKED ISSUE #1799: Add a "center" button to the component toolbar (horizontal centering via margin: auto)
### Context
A user gave this feedback during a Silex training: when an element is selected, it would be handy to center it horizontally in one click, straight from the toolbar that appears above the selection (the same little toolbar that has move / copy / delete).

### Current behavior
The component toolbar uses the GrapesJS default buttons (move, copy, delete, select parent). There is no quick "center" action, so to center a block a user has to open the style panel and set the margins by hand.

### Desired behavior
Add a button to the component toolbar that centers the selected element horizontally by applying `margin-left: auto; margin-right: auto` (i.e. `margin: 0 auto`) to it. Toggling it off should remove those two properties again.

### Where to start
- Commands are registered with `editor.Commands.add('silex:center-h', ...)`. A good home is a small dedicated module on the model of `editor/grapesjs/flex.ts` (which is also where the flex follow-up would live); `core-commands.ts` works too. The command operates on the currently selected component: `const cmp = editor.getSelected()`.
- **Write the style through the component API, not the Style Manager.** Read `cmp.getStyle()`, add or remove `margin-left` / `margin-right: auto`, then `cmp.setStyle(...)` (or `cmp.addStyle({...})`). GrapesJS puts this in a real CSS rule attached to the component's selector in the generated stylesheet — not an inline style. A working in-repo example of `setStyle` on a component is in `editor/publication-transformers.test.ts` (`comp.setStyle({ ... })`).
  - ⚠️ Do **not** use `editor.StyleManager.getSelected()` here. That returns whatever rule is active in the style panel (which may target nothing, or a different rule than the selected element) — it's only reliable from inside a `StyleManager.addType(...)` block like `background-image-asset`, not from a toolbar command.
- The component toolbar is currently the GrapesJS default (no custom `toolbar` is set in the repo). Extend the default component type's toolbar to add the button pointing to the new command. GrapesJS toolbar entries take an `attributes` object (with a `title`/aria-label) and a `command`.
- The editor uses Font Awesome icon classes (the GrapesJS default toolbar itself uses `fa fa-arrows-alt`, and Silex uses `fa-solid fa-*` throughout — see `editor/grapesjs/index.ts`). Reuse that convention for the button icon (e.g. `fa-solid fa-align-center`).
- Related: `editor/grapesjs/flex.ts` handles flex layout. Centering a **flex child** works differently (the parent's `justify-content`/`align-items`, or the child's `margin: auto` inside flex). For a first version, scope this to the `margin: 0 auto` block-centering case, which covers the common need. The flex case can be a follow-up.

### Acceptance criteria
- A "center" button shows in the toolbar above a selected element.
- Clicking it sets `margin-left: auto; margin-right: auto` on the element, via a proper CSS rule kept in the generated stylesheet (not inline style).
- Clicking it again removes them.
- The button has an accessible label (title / aria-label), and an icon consistent with the other toolbar icons.

### Testing note
`margin: 0 auto` only visibly centers a block that has a constrained `width`. Test on an element that has a width set, otherwise a full-width block will look unchanged (this is expected, not a bug).

### Notes
- Keep it clean: real CSS `margin: 0 auto`, no inline style, no lock-in.
- Horizontal centering only for this issue.

SparshM07:
Fixes #1799

Adds a Center horizontally toolbar action for selected components.

 - Applies `margin-left: auto` and `margin-right: auto` through the component API
 - Clicking the action again removes both properties
 - Uses a Font Awesome center-alignment icon with an accessible label

 Validation:
 - `corepack pnpm lint`
 - `corepack pnpm run build:client`
 - `corepack pnpm run build:server`

### DIFF
--- a/editor/grapesjs/center.ts
+++ b/editor/grapesjs/center.ts
@@ -1,0 +1,54 @@
+import { Editor } from 'grapesjs'
+
+const CENTER_COMMAND = 'silex:center-h'
+
+export default (editor: Editor) => {
+  editor.Commands.add(CENTER_COMMAND, () => {
+    const cmp = editor.getSelected()
+    if (!cmp) return
+
+    const style = cmp.getStyle()
+    const isCentered =
+        style['margin-left'] === 'auto' &&
+        style['margin-right'] === 'auto'
+
+    if (isCentered) {
+      const {
+        ['margin-left']: _marginLeft,
+        ['margin-right']: _marginRight,
+        ...newStyle
+      } = style
+
+      cmp.setStyle(newStyle)
+      return
+    }
+
+    cmp.setStyle({
+      ...style,
+      'margin-left': 'auto',
+      'margin-right': 'auto',
+    })
+  })
+
+  const defaultModel = editor.Components.getType('default').model as any
+  const originalDefaults = defaultModel.prototype.defaults
+
+  defaultModel.prototype.defaults = function () {
+    const defaults = originalDefaults.call(this)
+
+    return {
+      ...defaults,
+      toolbar: [
+        ...(defaults.toolbar || []),
+        {
+          attributes: {
+            class: 'fa-solid fa-align-center',
+            title: 'Center horizontally',
+            'aria-label': 'Center horizontally',
+          },
+          command: CENTER_COMMAND,
+        },
+      ],
+    }
+  }
+}
--- a/editor/grapesjs/center.ts
+++ b/editor/grapesjs/center.ts
@@ -30,13 +30,13 @@ export default (editor: Editor) => {
     })
   })
 
-  const defaultModel = editor.DomComponents.getType('default').model as any
-  const defaults = defaultModel.prototype.defaults
+  editor.on('component:selected', component => {
+    if (component.toolbar.some(item => item.command === CENTER_COMMAND)) {
+      return
+    }
 
-  defaultModel.prototype.defaults = {
-    ...defaults,
-    toolbar: [
-      ...(defaults.toolbar || []),
+    component.set('toolbar', [
+      ...component.toolbar,
       {
         attributes: {
           class: 'fa-solid fa-align-center',
@@ -45,6 +45,6 @@ export default (editor: Editor) => {
         },
         command: CENTER_COMMAND,
       },
-    ],
-  }
+    ])
+  })
 }
--- a/editor/grapesjs/center.ts
+++ b/editor/grapesjs/center.ts
@@ -1,3 +1,19 @@
+/*
+ * Silex website builder, free/libre no-code tool for makers.
+ * Copyright (c) 2023 lexoyo and Silex Labs foundation
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
 import { Editor } from 'grapesjs'
 
 const CENTER_COMMAND = 'silex:center-h'
--- a/editor/grapesjs/center.ts
+++ b/editor/grapesjs/center.ts
@@ -30,25 +30,21 @@ export default (editor: Editor) => {
     })
   })
 
-  const defaultModel = editor.Components.getType('default').model as any
-  const originalDefaults = defaultModel.prototype.defaults
+  const defaultModel = editor.DomComponents.getType('default').model as any
+  const defaults = defaultModel.prototype.defaults
 
-  defaultModel.prototype.defaults = function () {
-    const defaults = originalDefaults.call(this)
-
-    return {
-      ...defaults,
-      toolbar: [
-        ...(defaults.toolbar || []),
-        {
-          attributes: {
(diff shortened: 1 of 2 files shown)


lexoyo:
Thanks for picking this up so fast! I'll review it properly in the next few days.

One thing that might save you time meanwhile: in GrapesJS `Component.defaults` is a getter, not a data property, so you probably don't want to set `defaultModel.prototype.defaults`. Also please avoid the `as any` cast when possible, this is what prenvents the linter to warn you about this getter problem

The docs cover the supported way to do this: [Update Component Type](https://grapesjs.com/docs/modules/Components.html#update-component-type) shows how to modify an existing type while preserving its old defaults, and the [Component API](https://grapesjs.com/docs/api/component.html) documents the `toolbar` property if you'd rather set it per component on selection

SparshM07:
Thanks for the detailed feedback!!!
That makes sense — I hadn't realised "Component.defaults" is exposed via a getter, so setting "defaultModel.prototype.defaults" isn't the right approach. I'll rework the implementation using the approach described in the "Update Component Type" docs and avoid the "as any" cast as well. Thanks for pointing me in the right direction!

SparshM07:
Thanks for the detailed feedback — addressed in the latest commit.

I removed the Component.defaults mutation and the as any cast. The Center horizontally action is now added on component selection through the typed component.toolbar API, which preserves GrapesJS’s built-in toolbar actions.

  Validated with:

  corepack pnpm lint --fix
  corepack pnpm run build:client
  corepack pnpm run build:server

lexoyo:
Nice :)
Would you provide a screenshot too when you get a chance ? How does it look on your local environment ?

SparshM07:
I tried running the editor locally on Windows. Both the client and server builds succeed, but the dev server crashes while loading server/deploy/.silex.js because Node’s ESM loader rejects the Windows absolute path (C:\...) instead of a file:// URL. 

I’m looking into the local Windows workaround; I’ll share a screenshot once the editor starts...

SparshM07:
Tested locally on Windows — screenshot attached. The new Center horizontally action appears in the selected component toolbar and correctly centers images by toggling margin-left: auto and margin-right: auto.

For default full-width text blocks and inline buttons, the same margins are applied but may not cause a visible position  change because of their existing CSS display/width.

silex-center-toolbar

editor/grapesjs/center.ts
```diff
@@ -0,0 +1,50 @@
+import { Editor } from 'grapesjs'
```
lexoyo, line 17: Could you add the AGPL header at the top of the file? Every module in `editor/grapesjs/` starts with it, see `flex.ts` or `img.ts` for instance. Copying the block as is from one of them is fine

lexoyo:
Thanks for reworking it, and for pushing through the Windows setup to get that screenshot.

One question first, and please do not change any code for now. Now that you have the editor running, could you try this and tell me what you see? Select an element, click the new center button, then change the top margin of that same element from the style panel, then click the center button again.

I suspect the toggle stops working at that point, because the Margin property in Silex is a composite (see `editor/grapesjs/css-props.ts`), so GrapesJS writes it as the `margin` shorthand and clears the longhands. If that is confirmed, the fix may well be on my side in the style manager config rather than in your code, so do not start reworking anything. I just want to see it happen in the editor before I decide.

The only thing I actually need from you on this PR is the license header from my inline comment.

Unrelated to this PR, and not yours to fix either: the Windows failure you hit on the config file is a real bug on our side, and it blocks every Windows contributor. Could you open a separate issue with the exact error you got? That one is worth more to the project than this button

SparshM07:
Confirmed: after centering, changing Margin top through the Style panel serializes the styles as the margin shorthand and clears margin-left/margin-right. Clicking Center again is therefore treated as a new center action, so it does not toggle off. I have not reworked the code, as requested...

SparshM07:
I tested it. After centering an image, changing its Margin top in the Style panel, then clicking the horizontal-center button again, nothing visibly happens: the image remains centered. I did not change the code further..

lexoyo:
Thanks for running both tests — diagnosis confirmed: our Margin property is a composite, so GrapesJS serializes it as the `margin` shorthand (`margin: 10px 10px 10px 10px;`) and clears the longhands if there were some (`margin-left: auto; margin-right: auto;`). This is why the un-center does not work

To fix that, one last change: make the toggle shorthand-aware, inside your plugin. Expand the `margin` shorthand to its 4 values when checking `isCentered`, and when un-centering, clean both the longhands and the shorthand. That way the button works no matter what the style panel wrote.

Also: I opened #1824 for the Windows crash you hit, credited to you. Since you're on Windows, you're the best placed to work on it or verify a fix, if you feel like it.

lexoyo:
Update after more testing on your branch — I reproduced everything in the editor, and here is where the bugs you observed actually come from:

- The un-center issue is real, but harmless: it only happens when the style panel rewrites the margin between two clicks, the styles stay valid, and the style panel always lets you clean up. The fix belongs with the composite fix on our side, so no need for the shorthand change I asked for.
- The cases where the button visibly does nothing (inline-block, full-width elements) are not your code either: they are inherent to the `margin: auto` approach I specified in #1799. My spec, my bad.

So I'm merging as is: the normal path is clean and non-destructive, and your command + toolbar button are the right foundation to build on.

If you feel like continuing, pick whatever tempts you: #1824 (the Windows crash you found), or a follow-up on making this button context-aware (use the right centering technique depending on `display`, grey it out when it would do nothing). Zero obligation — congrats on your first contribution!

lexoyo:
Hello @SparshM07 
I just noticed a bug of the center button
When you select the body, it is visible alone at the top: 
image
Would you try and reproduce this? (I'm sorry I didn't take the time to do a proper issue)

SparshM07:
Yes, I'll try to reproduce this by selecting the <body> element and checking the center button behavior. I'll let you know what I find.

PR #1821, opened 2026-08-05
fix(editor): apply margin and padding per side in the style panel

lexoyo:
The Margin and Padding composite properties are now detached, so setting a single side writes that longhand instead of the margin/padding shorthand.


lexoyo:
Closing: with detached composite properties, grapesjs only writes the edited longhand and never clears a pre-existing margin/padding shorthand, so on existing websites the values it holds can no longer be removed from the UI.

PR #1825, opened 2026-08-14
 Fix Windows ESM config loading

### LINKED ISSUE #1824: [Windows] Server crashes at startup loading .silex.js — ESM loader rejects C:\ paths (even when no config file exists)
Reported by @SparshM07 while testing #1807 on Windows (thank you!): the dev server crashes at startup while loading `server/deploy/.silex.js`, because Node's ESM loader rejects the Windows absolute path. This blocks every Windows contributor at their first server start.

## Reproduction

On Windows, clone the repo, install, build, then start the server. It aborts with `ERR_UNSUPPORTED_ESM_URL_SCHEME` (protocol `c:`) while loading the default config file — whether or not a `.silex.js` file exists.

## Root cause

- `server/config.ts` resolves the default config path to a Windows absolute path (`C:\...\server\deploy\.silex.js`) and always attempts to load it.
- `common/silex-plugins/plugin.ts` (`getLocation`) runs the path through `new URL()`. On Windows this does not throw: the drive letter `C:` parses as a URL scheme, so the dynamic `import()` receives a `c:` protocol URL, which Node's ESM loader rejects (`Only URLs with a scheme in: file, data, node are supported`).
- The catch in `loadSilexConfig()` only treats `MODULE_NOT_FOUND` as "no config file, continue" — the ESM error doesn't match, so it is fatal even when no config file exists.

## Fix direction

- Convert filesystem paths to `file://` URLs with `pathToFileURL()` (from `node:url`) before the dynamic import. This belongs on the server side (`server/config.ts`, or a server-only branch in the plugin loader): `common/silex-plugins/plugin.ts` also runs in the browser, where the current URL handling is correct.
- A missing `.silex.js` must stay fatal. Do not extend the "config file not found" branch to `ERR_MODULE_NOT_FOUND`.
- Same treatment for user config files passed via the `CONFIG` env var, which go through the same code path (`loadUserConfig`).

Original report: https://github.com/silexlabs/Silex/pull/1807#issuecomment-5181202958

SparshM07:
Closes #1824
Convert server-side filesystem config paths to `file:` URLs before dynamic import, so Windows absolute paths do not become unsupported `c:` ESM URLs.
Also handles `ERR_MODULE_NOT_FOUND` when the default `.silex.js` file itself is absent, while preserving real module-loading failures.
 Tests: - `corepack pnpm test -- server/config.test.ts --runInBand`

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -36,6 +36,8 @@ import { ConnectorType } from '~/common/types.js'
 import { FsHosting } from './connectors/FsHosting.js'
 import { pathToFileURL } from 'node:url'
 
+const URL_SCHEME = /^[a-z][a-z0-9+.-]+:/i
+
 /**
  * Config types definitions
  */
@@ -149,7 +151,10 @@ export class ServerConfig extends Config {
       console.info('> Loading user config', this.userConfigPath)
       try {
         // Initiate the process with the config file which is just another plugin
-        await this.addPlugin(pathToFileURL(this.userConfigPath).href, {})
+        const userConfigLocation = URL_SCHEME.test(this.userConfigPath)
+          ? this.userConfigPath
+          : pathToFileURL(this.userConfigPath).href
+        await this.addPlugin(userConfigLocation, {})
       } catch (e) {
         throw new Error(`\nUser config file ${this.userConfigPath} error:\n\n\t${e.message}\n\n`, { cause: e })
       }
@@ -167,10 +172,7 @@ export class ServerConfig extends Config {
       await this.addPlugin(pathToFileURL(this.configFilePath).href, {})
     } catch(e) {
       // Check if the error is about the config file not found and not another module not found in the config file
-      if(
-        (e.code === 'MODULE_NOT_FOUND' && (!e.requireStack || !e.requireStack.find(path => path === this.configFilePath))) ||
-        (e.code === 'ERR_MODULE_NOT_FOUND' && e.url === pathToFileURL(this.configFilePath).href)
-      ) {
+      if(e.code === 'MODULE_NOT_FOUND' && (!e.requireStack || !e.requireStack.find(path => path === this.configFilePath))) {
         console.info('> /!\\ Config file not found', this.configFilePath)
       } else {
         throw new Error(`Error in config file ${this.configFilePath}: ${e.message}`, { cause: e })
--- a/server/config.ts
+++ b/server/config.ts
@@ -24,7 +24,7 @@
  * @class default config for Silex server
  */
 
-import { Config, Plugin } from '~/common/silex-plugins/index.js'
+import { Config } from '~/common/silex-plugins/index.js'
 import { join, resolve } from 'path'
 import { CLIENT_CONFIG_FILE_NAME } from '~/common/constants.js'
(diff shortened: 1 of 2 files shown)


lexoyo (commented):
Thanks @SparshM07, I ran your branch on Linux: the tests pass, the server starts, and the ESM path conversion is right.

Two changes requested inline. One of them is my fault, my issue asked for something wrong, sorry.

Your branch also still contains bits of your previous contribution, because it starts from `fix-1799` instead of `main`. To remove them:

```sh
git pull --rebase
git push --force-with-lease
```

If you can, paste the server startup logs from your Windows machine. Not blocking.

server/config.ts
```diff
@@ -148,7 +149,7 @@ export class ServerConfig extends Config {
       console.info('> Loading user config', this.userConfigPath)
       try {
         // Initiate the process with the config file which is just another plugin
-        await this.addPlugin(this.userConfigPath, {})
+        await this.addPlugin(pathToFileURL(this.userConfigPath).href, {})
```
lexoyo: `SILEX_SERVER_CONFIG` can already be a URL (`file://...`), and `getLocation()` used to handle that. Here it would be turned into a broken path. Please convert only when the value is not already a URL.

Careful with the test you use for this: a Windows path like `C:\...` looks like a URL with the scheme `c`, which is the bug we are fixing. Requiring at least two characters before the colon avoids it, for example `/^[a-z][a-z0-9+.-]+:/i`.

server/config.ts
```diff
@@ -163,10 +164,13 @@ export class ServerConfig extends Config {
...
     try {
       // Initiate the process with the config file which is just another plugin
-      await this.addPlugin(this.configFilePath, {})
+      await this.addPlugin(pathToFileURL(this.configFilePath).href, {})
     } catch(e) {
       // Check if the error is about the config file not found and not another module not found in the config file
-      if(e.code === 'MODULE_NOT_FOUND' && (!e.requireStack || !e.requireStack.find(path => path === this.configFilePath))) {
+      if(
+        (e.code === 'MODULE_NOT_FOUND' && (!e.requireStack || !e.requireStack.find(path => path === this.configFilePath))) ||
+        (e.code === 'ERR_MODULE_NOT_FOUND' && e.url === pathToFileURL(this.configFilePath).href)
```
lexoyo: Sorry, this part of my issue was wrong. A missing config file must stay fatal.

Please remove this change.

SparshM07:
Thanks for reviewing! I’ve updated the branch as requested and removed the previous contribution commits by rebasing/pushing the branch. The PR checks are now all passing with no conflicts.

I’ll also share the Windows server startup logs shortly.

lexoyo:
Awesome @SparshM07 :1st_place_medal: 

I ran your branch on Linux before merging: tests pass, the server starts, SILEX_SERVER_CONFIG works both as a path and as a file:// URL, and a missing config file still fails the same way as before :+1: 

Thank you! See you for the next one? :))

SparshM07:
Thank you so much! Really appreciate you taking the time to review and test it on Linux.
I definitely learned a lot from this contribution and from your feedback throughout the process. Looking forward to the next one! 
See you around! :)

PR #1829, opened 2026-08-20
fix(publication): omit empty lang and fall back to page name for title

### LINKED ISSUE #1734: Make page title and html lang automatic for published sites
Discussed on the community forum (https://community.silex.me/d/291). Opening it here so it does not get lost.

## Problem

Looking at a number of sites published with Silex, one small thing keeps coming back: many ship without a page `<title>` and without a `lang` attribute on the `<html>` tag.

Both matter more than they look:
- A missing `<title>` means broken browser tabs, weak SEO, and a poor link preview when the page is shared.
- A missing `lang` means a screen reader can mispronounce the whole page. Setting it is a basic accessibility requirement (WCAG 3.1.1, Language of Page).

Today nothing in the flow nudges you to set them, so it is easy to publish a site without ever thinking about it.

## Idea (from the forum, comments included)

- Better defaults so a fresh page already has a title and a language set, and the common case is correct out of the box.
- A light pre-publish check that warns when a page has no title or no lang, without blocking publication.

## Why now

Accessibility and clean output are part of where Silex is going, and this is a small change with a real impact on every published site.

SparshM07:
>What does this PR do?

Fixes #1734.

- Omit the `lang` attribute when no language is configured instead of generating `lang=""`.
- Fall back to the page name for `<title>` when no title is configured.
- The fallback happens at publish time and is not persisted in page settings.
- No default language is guessed.
- Publishing is not blocked when these settings are empty.

>Testing

- Added tests covering the requested cases.
- All 5 new tests pass.
- Full editor test suite passes.
- ESLint passes.
- Tested the published output in the browser.

### DIFF
--- a/editor/grapesjs/PublicationManager.test.ts
+++ b/editor/grapesjs/PublicationManager.test.ts
@@ -1,0 +1,63 @@
+/*
+ * @jest-environment jsdom
+ */
+
+import { expect, jest, describe, it, beforeEach } from '@jest/globals'
+import grapesjs, { Editor } from 'grapesjs'
+import { WebsiteSettings } from '~/common/types'
+import { PublicationManager } from './PublicationManager'
+
+// Prevent lit-html from being imported (it is a peer dependency and breaks the tests)
+jest.mock('lit-html', () => ({}))
+
+describe('PublicationManager html output', () => {
+  let editor: Editor
+
+  beforeEach(() => {
+    /* @ts-ignore */
+    editor = grapesjs.init({
+      headless: true,
+      storageManager: { autoload: false },
+    })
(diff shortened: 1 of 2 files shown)


lexoyo:
Merged, thank you. The tests you added made this easy to review, and I checked the published output on a real site: no more empty lang, and the page name as the title.

If you want another one, have a look at the issues with the good first issue label.

SparshM07:
Thank you so much! 😄 Really glad the tests and implementation made the review straightforward. I really appreciate all the guidance and feedback throughout these contributions.

And absolutely — I'll have a look at the good first issues for the next one!

PR #1833, opened 2026-08-22
feat(editor): warn about images missing alt text

SparshM07:
>>Summary
Adds a non-blocking pre-publish accessibility warning when a page contains images without an `alt` attribute.

- Reports the number of images missing `alt` per page
- Links the warning to the first affected image
- Treats explicit `alt=""` as handled for decorative images
- Adds focused unit tests

>>Testing

- `PublicationManager.test.ts`: 7 tests passed
- ESLint passed

### DIFF
--- a/editor/grapesjs/PublicationManager.test.ts
+++ b/editor/grapesjs/PublicationManager.test.ts
@@ -3,7 +3,7 @@
  */
 
 import { expect, jest, describe, it, beforeEach } from '@jest/globals'
-import grapesjs, { Editor } from 'grapesjs'
+import grapesjs, { Component, Editor } from 'grapesjs'
 import { WebsiteSettings } from '~/common/types'
 import { PublicationManager } from './PublicationManager'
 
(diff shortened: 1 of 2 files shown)


lexoyo (commented):
Thanks, this is clean and it hooks in at the right place.

Two things.

**Use the existing `seo-validation` group.** These warnings are already a mix of SEO and accessibility, not two separate families. The `lang` check right above yours says so itself. A second group splits the same kind of warning into two blocks for the user.

**One notification per image, not one per page.** Yours points to the first image only, so the user fixes it, publishes to find the next one, and every publish runs a build on the forge. One per image lets them fix everything in one pass. The notification id is what deduplicates, so it needs to be unique per image.

No need to cap the number yourself, the plugin already does.

The rest is good. Treating `alt=""` as intentional is the right call.

As usual, please test it yourself in a browser, and send me a screenshot if possible :)

lexoyo:
One more thing, about my second point.

One notification per image may be slow on a page with many images. The plugin saves the whole list to localStorage on every add, so the cost grows with each one. Its display limit does not help here, it only truncates what is shown.

So please try it with a lot of images and tell me if it feels slow. If it does, we will change the notification a bit rather than drop the idea.

Issue #1678, opened 2026-03-29
Missing CSS transition properties in the editor


lexoyo:
## Description

The CSS transitions tab in the editor is missing several commonly used properties.

## Missing properties

- `color`
- `border-color`
- `padding`
- `margin`
- `top`, `bottom`, `left`, `right`
- `cubic-bezier(n,n,n,n)` timing function

## Use case

- `cubic-bezier` enables bouncy effects and smoother animations
- Positional properties (`top`, `left`, etc.) are needed for animated layouts

## Source

Reported by ceubri on the community forum: https://community.silex.me/d/186

lexoyo:
CSS properties in the Style Manager are added in editor/grapesjs/css-props.ts, with editor.StyleManager.addProperty('extra', { ... }) (see the existing ones there, around lines 707 to 830, added to the "Effects" sector). Adding transition support means adding the transition properties the same way: transition-property (with options such as color, border-color, padding, margin, top, right, bottom, left), transition-duration, transition-timing-function (including a cubic-bezier option), and transition-delay.

Acceptance: the transition fields appear in the Style Manager, you can set them on an element, and they show up in the exported CSS. Please check with a simple case, for example a color transition on hover.

dyk1454683243-sudo:
I'll finish the leftover transition controls from #1838.

#1838 already added `color`, `border-color`, `filter`, `backdrop-filter`, and `visibility` to `transition-property`, and kept the GrapesJS three-part stack (property / duration / timing-function). Still missing from this issue:

- `padding`, `margin`, `top`, `bottom`, `left`, `right` on the property dropdown
- a usable `cubic-bezier(...)` timing function

GrapesJS select options need a concrete `value` string, so I will add a small set of named cubic-bezier presets (including a common bounce / overshoot curve) rather than a free-form bezier editor. A custom-curve control would be a different input type and is out of scope here.

Not reopening #1835 — `text-underline-offset` and `transform-origin` stay as merged. Separate PR from the MCP work (#1866 / #1868).

dyk1454683243-sudo:
PR up: https://github.com/silexlabs/Silex/pull/1872 — leftover transition-property options (`padding`, `margin`, `top`, `bottom`, `left`, `right`) plus named `cubic-bezier(...)` timing-function presets after #1838.

PR #1838, opened 2026-08-26
feat(editor): add text-underline-offset, transform-origin and more transition properties

### LINKED ISSUE #1835: Style panel: add text-underline-offset, transform-origin, and filter in transitions
### Context

A user asked for these three options in the forum: https://community.silex.me/d/313-interface-css-ajout

They are missing from the style panel today. The three changes are in the same file: `editor/grapesjs/css-props.ts`

### 1. Add `text-underline-offset` (Typography)

The panel has "Text decoration" with line, style, color and thickness. There is no offset, so you cannot move the underline away from the text.

Add it as its **own** property in the `typography` sector, next to the existing "Text decoration" block.

Careful: `text-underline-offset` is not part of the `text-decoration` shorthand. Do not add it inside the existing composite property. If you do, Silex writes an invalid rule like `text-decoration: underline solid red 2px 4px` and the underline disappears.

Units: px, em, rem, %. Fixed values: auto, inherit, initial, revert, unset.

### 2. Add `transform-origin` (Effects)

`transform` is already in the "Effects" sector, but you cannot change the point the element rotates or scales around.

Add `transform-origin` in the `extra` sector, next to `transform`. The value has one or two parts, x and y. Each part is a keyword (left, center, right / top, center, bottom) or a length. A composite property with two sub-properties fits this.

### 3. Allow transitions on more properties, including `filter` (Effects)

The "Transition" property has a "Property" dropdown. Today it lists only: all, width, height, background-color, transform, box-shadow, opacity. This list comes from GrapesJS, not from Silex.

`filter` is missing, but Silex adds the `filter` property itself (see `addProperty('extra', { extend: 'filter' })` in `editor/grapesjs/index.ts`). So today you can blur an element, but you cannot animate the blur.

Add at least `filter`. Please also add `color`, `border-color`, `backdrop-filter` and `visibility`.

One way to do it: remove the property and add it again with your own list, like the file already does for `display` (`removeProperty('general', 'display')` then `addProperty(...)`). Keep the same three sub-properties: property, duration, timing function.

### Before you open the pull request

Try each new setting in the editor, with the mouse, in a real browser. Post one screenshot per setting, so we can see it working:

- an underline moved down by `text-underline-offset: 10px`. Please use a wavy red thick underline for this one, so we can see that the other text decoration settings still work
- an element with rotate `45deg` turning around its top left corner
- `filter` in the transition dropdown, and a blurred element on hover with a `2s` transition on `filter`

Then save, reload the page in your browser, and select the same elements again. The values must still be in the panel

ayush2115:
## Related discussion

Closes #1835 — forum thread: https://community.silex.me/d/313-interface-css-ajout

One PR for the three changes because the issue groups them, and they are all in the same file (`editor/grapesjs/css-props.ts`).

## What changed

**1. `text-underline-offset` in Typography**, directly after "Text decoration". Units `px, em, rem, %`, fixed values `auto, inherit, initial, revert, unset`.

It is its **own** property, not a fifth sub-property of the existing composite. `text-underline-offset` is not part of the `text-decoration` shorthand, so adding it there would make GrapesJS join it into `text-decoration: underline solid red 2px 4px`, which is invalid — and CSS requires the whole declaration to be dropped, so the line, the style and the colour would disappear along with the offset.

**2. `transform-origin` in Effects**, directly after `transform`. A composite with an X, a Y and a Z part. X and Y take a length, a percentage, or the keywords of their own axis (`left/center/right` and `top/center/bottom`). Z takes a length only — no percentage, which is invalid in that slot and would void the whole declaration, and no keyword, since none is allowed there.

It carries `toStyle`/`fromStyle`. That is also what makes Z safe: CSS only accepts a Z after an X and a Y, and `toStyle` already writes both in every case, so appending a Z can never produce an invalid value.

`toStyle` also writes a CSS-wide keyword (`inherit`, `initial`, `revert`, `unset`) on its own, and lets it win over the axes. Since both axes are otherwise always written, picking one on an axis would give `transform-origin: inherit center` — invalid, and browsers drop the whole declaration, so the setting would silently do nothing. The keywords are offered on X only, where `fromStyle` reads a lone token, so a `transform-origin: inherit` written outside Silex shows up in the panel and can be cleared rather than looking unset and being overwritten by the next edit.

Without `toStyle`/`fromStyle` the property is broken on reload: GrapesJS joins composite sub-values positionally and drops the empty ones, so a Y set on its own is written as `transform-origin: top` and read back into the **X** field — where `top` is not a valid value, so it is silently discarded and the setting is lost. The property now always writes both parts, the empty one as `center` (the value the browser uses anyway), and parsing accepts the forms CSS allows but Silex does not write: a single token, and `top left` as well as `left top`.

**3. More properties in the Transition dropdown.** The list came from GrapesJS and only had `all, width, height, background-color, transform, box-shadow, opacity`. `filter` was missing even though Silex registers the `filter` property itself (`editor/grapesjs/index.ts`), so you could blur an element but not animate the blur. The property is removed and re-added with `filter`, plus `color`, `border-color`, `backdrop-filter` and `visibility`. Same three sub-properties (property, duration, timing function), same defaults as GrapesJS.

The `at:` indices of the properties that follow the two new ones are shifted by one, so the new properties sit next to the ones they relate to rather than at the bottom of the sector.

`editor/grapesjs/css-props.test.ts` is new: panel order, the CSS actually written, and the read-back after a reload — including a regression test for the `transform-origin` axis loss and one asserting `text-underline-offset` did **not** end up inside the `text-decoration` composite.

## How to test

`pnpm install && pnpm build && pnpm start`, then in the editor:

### 1. `text-underline-offset: 10px` on a wavy red 3px underline

Text decoration line `underline`, style `wavy`, colour `#e01b24`, thickness `3px` — all four still work — and the offset pushes the wave down.

text-underline-offset

### 2. `rotate 45deg` around the top left corner

`transform: rotateZ(45deg)` with Transform origin X `left`, Y `top`. The box pivots on its top-left corner.

transform-origin

And the Z axis, added after review. Two boxes, identical `rotateX(45deg)`, the second one with `transform-origin: center center 100px` — the Z offset alone moves it 71px down (measured `y=133` vs `y=204`). No `perspective` needed:

transform-origin Z

After saving and reloading, `transform-origin: center center 100px` is still there and the panel reads back `x=center y=center z=100px`:

transform-origin Z after reload

A CSS-wide keyword typed into X writes `transform-origin: inherit` on its own, and after a save and reload it is still shown, highlighted, and clearable:

css wide keyword after reload

Every value the property can emit was checked in a browser against a marker rule — `inherit`, `initial`, `revert`, `unset`, `left top`, `center top` and `center center 30px` all apply; `inherit center`, which the first version emitted, is rejected outright.

### 3. `filter` in the transition dropdown

Transition layer `filter 2s ease`:

transition on filter

The Property dropdown options (a native `<select>` popup cannot be captured in a screenshot, so the list is rendered inline here):

transition property options

`filter: blur(2px)` on the `:hover` state of the same element:

hover filter blur

And the transition actually running on the exported page — no hover, 600 ms into the 2 s transition, and finished:

| no hover | 600 ms | done |
|---|---|---|
|  |  |  |

### 4. Save, reload the browser, select the same elements again

The values are still in the panel:

| `text-underline-offset` | `transform-origin` | `transition` |
|---|---|---|
|  |  |  |

The CSS after reloading:

```css
#para { text-decoration: underline wavy #e01b24 3px; text-underline-offset: 10px; }
#box  { transform: rotateZ(45deg); transform-origin: left top; }
#deep { transform: rotateX(45deg); transform-origin: center center 100px; }
#blur { transition: filter 2s ease; }
#blur:hover { filter: blur(2px); }
```

The screenshots were taken by driving the editor in a real Chromium. Every value visible in the panel — the four text-decoration fields, the offset, the transform layer, both transform-origin axes, the transition layer and the hover blur — was set through the controls; only the demo page content and its layout CSS (sizes, colours, margins) were created through the API. The images live on a [separate branch](https://github.com/ayush2115/Silex/tree/screenshots/issue-1835) of my fork so they stay out of this diff.

### Checks

`pnpm build`, `pnpm lint` and `pnpm test` are green. `server/connectors/FsStorage.test.ts` fails on my machine on Node 26 (`rmdirSync` lost `options.recursive`); it fails the same way on `main` without this change, and CI pins Node 24.

## Notes

- **`backdrop-filter` needs `-webkit-` on Safari 9–17**, both as a property and inside `transition-property`. Silex has no prefixing pipeline, so a `backdrop-filter` transition is dead on those versions. Out of scope here.
- **`text-underline-offset` with `%` is ignored on Safari before 18.2** while working in Chrome and Firefox. `%` is in the units list because the issue asks for it; `em` is the safe choice.
- **`visibility` transitions need one endpoint to be `visible`** — `hidden` → `collapse` never animates. Spec behaviour, not something this PR can fix.
- **Pre-existing, not touched here:** the other composites in this file collapse partial values the same way `transform-origin` would have. Setting only the decoration colour writes `text-decoration: red`, which comes back in the **line** sub-property after a reload; `outline: red` comes back empty; and the `overflow` composite has no `property` key on its sub-properties, so they are named `Overflow-X`/`Overflow-Y` (`Property.initialize` derives the name from the label without lowercasing). Worth a separate issue — flagging it so that a partial `text-decoration` check while reviewing screenshot 1 does not look like a regression from this PR. Happy to open one, or to fix them here if you prefer.
- #1678 overlaps point 3 and additionally asks for `transition-delay` and `cubic-bezier`, which #1835 explicitly rules out ("keep the same three sub-properties"), so this PR does not close it.


lexoyo (changes requested):
@ayush2115 this is great work!!

Do you think you could add the Z axis as well? Units can be defined in the objects, or am I missing something?

Yes please, open that issue about the other composites :pray: 

By the way, I didn't know about the toStyle/fromStyle API in grapesjs, I believe I should have used it for `object-position` in `editor/grapesjs/img.ts`, where the nine keyword pairs are written by hand today. Thanks for the tip :+1: Have you used grapesjs before contributing to Silex?

Talk to you soon, bye!

ayush2115:
Thanks! Z added, I had the units wrong in my head. Each sub property carries its own, so Z takes `px` / `em` / `rem` and simply never offers a percentage, which is what had made me skip it.

CSS only accepts a Z after an X and a Y, and `toStyle` already wrote both in every case, so appending Z can't produce an invalid value. Screenshot in the description: two boxes with the same `rotateX(45deg)`, the second one with `transform-origin: center center 100px` sits 71px lower , no `perspective` needed.

While testing that I found a bug in my own property and fixed it here too. Because `toStyle` always writes both axes, picking `inherit` / `initial` / `revert` / `unset` on an axis produced `transform-origin: inherit center`, which is invalid, so browsers dropped the whole declaration and the setting silently did nothing. It now writes the keyword on its own and lets it win over the axes. I checked every value the property can emit against a marker rule in a browser, `inherit center` is the only one that gets rejected. The keywords are offered on X only, where `fromStyle` reads a lone token, so a `transform-origin: inherit` written outside Silex shows in the panel and can be cleared instead of looking unset and being overwritten by the next edit.

All three settings still survive a save and a reload, screenshots in the description. 32 tests on this file now.

Opened #1847 for the other composites.

This PR was my first time with GrapesJS. The composite behaviour isn't really documented, so I ended up reading `PropertyComposite` / `InputNumber` to work out what `toStyle` / `fromStyle` actually do.

lexoyo (approved):
Great, thanks a lot for the Z axis, for the keyword fix, and for opening #1847

Reading `PropertyComposite` to work out what `toStyle` and `fromStyle` really do is exactly what I ended up doing too. I will use them for `object-position` in `img.ts`

PR #1843, opened 2026-08-27
feat(editor): add <time> tag and datetime trait in element settings

### LINKED ISSUE #1692: Add <time> element to tag name settings
Allow the addition of the <time> tag in Silex.

The `<time>` HTML element is missing from the tag name selector in element settings. Same family as #1679 (missing `<hr>`) — could be addressed together.

Image

### Related
- Roadmap: https://roadmap.silex.me/posts/59
- MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time

britsync07-prog:
## Description
Closes #1692

This PR adds the <time> tag to the tag name selector in element settings and adds support for the datetime attribute trait when <time> is selected.

### Changes
- Added 'TIME' to the **Content grouping** category in 	agCategories within editor/grapesjs/semantic.ts.
- Registered datetime-trait and added conditional rendering so the datetime input is displayed only when the selected element's tag is <time> (matching the behavior of or-trait on <label>).
- Added unit tests in editor/grapesjs/semantic.test.ts to test trait registration, <time> tag selection, and datetime attribute rendering in exported HTML.

### Verification
- [x] 
px eslint editor/grapesjs/semantic.ts editor/grapesjs/semantic.test.ts passes with 0 errors
- [x] 
ode --experimental-vm-modules node_modules/jest/bin/jest.js editor/grapesjs/semantic.test.ts passes (3/3 tests)
- [x] pnpm run build:client passes
- [x] pnpm run build:server passes


lexoyo (changes requested):
Thanks, this fits the `for-trait` shape well.

Can you make the two traits share one implementation? They only differ by the attribute name, the tag, the label and the placeholder, and more are coming (`cite` on `<blockquote>`, `open` on `<details>`).

Please also remove `semantic.test.ts`. The three tests would stay green even if the field never worked, and a test that protects nothing is still code to maintain. Check by hand instead that the field shows only on `<time>`, and that typing writes the attribute (and take a screenshot to post in the PR)

britsync07-prog:
@lexoyo Updated! Refactored the or and datetime traits into a shared generic conditionalTraits registry and unified trait renderer in editor/grapesjs/semantic.ts. This makes adding future tag-specific attribute traits (like cite on <blockquote> and open on <details>) as simple as adding an entry to the registry. Also removed semantic.test.ts as requested.

lexoyo:
Nice, the registry is exactly what I had in mind

One thing I noticed while testing: `setAttributes` replaces the whole attributes object, so typing in the Datetime or For field drops the other attributes of the element. On a `<time id="t1" title="hello" data-foo="bar">` only `datetime` and `id` are left. `addAttributes` merges instead. It comes from the old `for-trait` code, not from your change, but since you are in there, could you switch the two calls?

Could you also post a screenshot of the Datetime field on a `<time>` element?

britsync07-prog:
@lexoyo Done! Switched the shared conditional-trait handler from `setAttributes` to `addAttributes`, so typing in Datetime/For now merges instead of replacing - other attributes (`id`, `title`, `data-*`) survive. Since both traits share the one registry renderer, this single call site covers both. ESLint clean, no new `tsc` errors in `semantic.ts`.

I couldn't capture a browser screenshot from this environment - could you please confirm visually that the Datetime field shows only on `<time>` and typing preserves the other attributes?

PR #1844, opened 2026-08-30
fix(editor): allow leaving preview mode with Escape key

### LINKED ISSUE #1841: Preview mode cannot be left with the keyboard
### What happens

There is no way out of preview mode with the keyboard. Escape was pressed twice, nothing happened.

The only way out is the `gjs-off-prv` icon: 24 px, no label, in the very corner of the screen at (0,0). In the desktop app there is no browser tab to fall back on, so a user who does not find that icon is stuck in the preview.

### Expected

Escape leaves the preview.

britsync07-prog:
Resolves #1841

### Description
Pressing the `Escape` key now properly exits preview mode and returns to the editor, matching user expectations and preventing users from getting trapped in preview mode (especially in desktop/fullscreen contexts).

### Changes
- Updated `escapeContext()` in `editor/grapesjs/keymaps.ts` to check for active preview mode (`editor.Commands.isActive('preview')`) and trigger `editor.stopCommand('preview')`.
- Updated the keydown event listener in `keymapsPlugin` to handle `Escape` when in preview mode and safely attach listeners across both the outer document and canvas iframe.
- Added type/function safety guards to `isTextOrInputField()` in `editor/utils.ts`.
- Added unit test suite in `editor/grapesjs/keymaps.test.ts` verifying default keymaps registration, Escape key behavior during preview mode, and modal dismissal.

### Testing & Verification
- ESLint: 0 errors, 0 warnings (`npx eslint editor/grapesjs/keymaps.ts editor/grapesjs/keymaps.test.ts editor/utils.ts`)
- Unit Tests: 7/7 DOM test suites passed (45/45 tests passing) (`jest --selectProjects dom`)
- Client Webpack Build: Compiled successfully (`pnpm run build:client`)
- Server Build: Compiled successfully (`pnpm run build:server`)


lexoyo:
Thanks!

One thing to check before I read the code: I think the key events of the canvas already reach the main document, GrapesJS forwards them. If so, adding a preview case in `escapeContext()` may be all we need, and the rest can go.

As usual, please test it yourself in a browser, and send me a screenshot if possible :)

lexoyo (changes requested):
Thanks!

One thing to check before I read the code: I think the key events of the canvas already reach the main document, GrapesJS forwards them. If so, adding a preview case in `escapeContext()` may be all we need, and the rest can go.

As usual, please test it yourself in a browser, and send me a screenshot if possible :)

britsync07-prog:
@lexoyo Updated! You're completely right - GrapesJS forwards key events from the canvas to the main Keymaps handler. I have streamlined the implementation so it only adds the active preview check (editor.Commands.isActive('preview')) inside escapeContext() and removed all extra listeners. The diff is now minimal and clean.

lexoyo:
Thanks, this is much smaller now. I tried it and it works

Could you remove `keymaps.test.ts`? It replaces `Commands.isActive` with a mock, so it stays green even if the feature breaks. The `?.` in `keymaps.ts` and the guard in `isTextOrInputField` are only there for that test, they can go with it

britsync07-prog:
@lexoyo Done! Removed `keymaps.test.ts` and reverted the test-only guards - the `?.` in `keymaps.ts` and the null guard in `isTextOrInputField`. The branch diff vs `main` is now just the `escapeContext` preview check plus its docstring line. Ran ESLint on both touched files, clean.

PR #1848, opened 2026-09-02
feat(editor): add basic funtionality for localization + sample strings

pbouda:
## Related discussion

This adds first support for basic localization to the front-end, we discussed via mail.

## What changed

Lays the groundwork for localizing Silex's own editor UI, reusing the editor.I18n/addMessages() mechanism already proven in grapesjs-advanced-selector instead of introducing a second i18n library.

### Core infra
- t(editor, key, vars?) / initI18n() / resolveLocale() in editor/src/i18n, backed by plain JSON locale files (editor/src/locales/en-US.json, fr-FR.json) loaded via require.context
- Locale resolution: ?lang= → localStorage → navigator.language → en-US, with a bare-language fallback (fr → fr-FR) for callers that don't yet send full BCP-47 codes
- A language switcher in Site Settings → General, persisted across sessions
- npm run i18n:extract: scans t(editor, '...')/msg('...') call sites and syncs them into en-US.json (--check for CI, --prune to remove stale keys)

### Proof-of-concept migration
10 representative strings wrapped end to end (dialog title, buttons, confirmation prompt, tooltip, block label, HTML-bearing paragraph, an interpolated notification), plus a complete French translation of all 10.

Scope: intentionally small by design. This shows the pattern; migrating the more/rest of the UI's, back-end and desktop app strings is separate follow-up work. Also the plugins' localization will happen separately in a new PR.

## How to test

Start the Silex editor, open "Settings" and change the editor language to French "fr-FR".


editor/src/i18n/index.ts
```diff
@@ -0,0 +1,103 @@
...
+ * reusing the getTranslation()/addMessages() pattern already proven in
+ * grapesjs-advanced-selector, instead of a second JS i18n library.
+ * Locale files are plain JSON identity maps (English string -> translation),
+ * one full, standalone file per BCP-47 locale, loaded as-is with no
+ * merge/overlay step.
+ */
+
+import { Editor } from 'grapesjs'
+
+const localesContext = require.context('../locales', false, /\.json$/)
```
lexoyo, line 29: This runs when the module loads, and `require.context` only exists under webpack, so `editor/grapesjs/PublicationManager.test.ts` does not run any more:

    ReferenceError: require is not defined

The PR already turns on `resolveJsonModule`, so importing the two JSON files directly would work in both jest and webpack. Would that work for you?

lexoyo (changes requested):
Thanks for this @pbouda :pray: 

Would any of those be a problem for Weblate? Let me know what you think?

1. It is preferable to keep short locale codes, `en.json` and `fr.json`. GrapesJS stores its own labels under `en` and ships `fr.js`, `es.js`, `de.js` for the standard UI. Also on our side, long locale codes would require a fair amount of changes
2. I believe short symbolic keys are easier than the full text in our case, like `pages.delete.confirm`. Some keys are full paragraphs, and one is a block of HTML with links, what do you think?
3. I'm not sure if GrapesJS views listen to `i18n:locale`, so switching the locale at runtime may not re-render the style manager, the traits, the blocks or the panels. Did you test it? The select could simply reload with `?lang=` if it's easier. For now we default to the browser language, and let `?lang=` win. The dashboard already decides the language and passes it in the URL
4. We should move the select box in the settings to somewhere in the UI as the settings are at the site level. But I guess you wanted to test a dropdown, it's ok I can move it in another PR maybe. I'll ask on the forums what users want

pbouda:
Here is my thoughts:

>     1. It is preferable to keep short locale codes, `en.json` and `fr.json`. GrapesJS stores its own labels under `en` and ships `fr.js`, `es.js`, `de.js` for the standard UI. Also on our side, long locale codes would require a fair amount of changes

I understand. I just suggest to use codes with region, as this is the most complete solution that projects normally will use in the long run anyway. E.g. Swiss German will already require it, there is some significant differences to standard german (ß vs. ss among others), and the Swiss users will notice. If you really think that's a major problem then we can keep the simple language codes. What other changes would be required?


>     2. I believe short symbolic keys are easier than the full text in our case, like `pages.delete.confirm`. Some keys are full paragraphs, and one is a block of HTML with links, what do you think?

Using English as keys is the better approach normally, as this allows to work with more or less normal strings in the code and is much easier to maintain in the rest of the translation workflow, as the tools are prepared for that. The "keys" are not visible as keys anyway, you would not edit the .po files directly. In the weblate interface the translators will translate from English strings to the target. HTML and other markup is normal there, the translators should be able to handle it.


>     3. I'm not sure if GrapesJS views listen to `i18n:locale`, so switching the locale at runtime may not re-render the style manager, the traits, the blocks or the panels. Did you test it? The select could simply reload with `?lang=` if it's easier. For now we default to the browser language, and let `?lang=` win. The dashboard already decides the language and passes it in the URL

Yes, the `lang=` parameter works already, from the dashboard also. Maybe it's indeed easiest to reload with the new `lang=` parameter, if that's OK with you. We can add this when we know the final place for the language select.


>     4. We should move the select box in the settings to somewhere in the UI as the settings are at the site level. But I guess you wanted to test a dropdown, it's ok I can move it in another PR maybe. I'll ask on the forums what users want

Yes, I just added it for testing, we can move it anywhere. If you know what users want, let me know.

editor/src/i18n/index.ts
pbouda: Yes, that would work, but I think we would loose the feature that locales are auto-discovered from the JSON files. We can add this to the documentation, when adding new languages what to do. Is that ok? I think we could also add a jest polyfill to support the require.context, but maybe that's overcomplicating things.

lexoyo:
Thanks for your answers @pbouda :pray:

1. About locale codes: GrapesJS uses ISO 639-1 codes (`fr`, `de`) and ships its translations with these names. The Weblate docs also say they prefer two letter codes, with a region only for variants like `de_CH` or `pt_BR`. With `fr-FR` everywhere we would have to map all the GrapesJS locales. So I would keep `en` and `fr`, and add a region only when we need a variant like Swiss German. What do you think?

2. OK for English strings as keys. GrapesJS recommends putting a plugin's strings under the plugin name (see the [I18n docs](https://grapesjs.com/docs/modules/I18n.html)). So we could put ours under `silex`, and each plugin in this repo under its own name, like some of them already do. We will change the other plugins to do the same.

I guess the JSON files can stay flat, for example `editor/src/locales/fr.json`:

```json
{ "Cancel": "Annuler" }
```

and the namespace is added in the code:

```ts
editor.I18n.addMessages({ fr: { silex: fr } })
```

Then Weblate has one component per plugin, each with flat files. What do you think?

Also I'm not sure `editor.I18n.t()` will find keys with a dot inside, because it uses dots for nested keys. For example `t('silex.Are you sure? This cannot be undone.')` may look for `silex` > `Are you sure? This cannot be undone` > `''`. Maybe `t()` should read the `silex` messages directly?

3. and 4. Let's choose the language in the dashboard only. The SaaS dashboard already has an EN / FR switch and passes `?lang=` to the editor, and the new Desktop dashboard will do the same. So let's remove the select and the localStorage from this PR, the editor can read `?lang=` or use the browser language

editor/src/i18n/index.ts
lexoyo: Let's import them explicitly

PR #1849, opened 2026-09-02
fix(pages): confirm before deleting a page from the pages panel

### LINKED ISSUE #1839: Deleting a page asks nothing, and the trash icon is in the middle of the row
### What happens

Clicking the trash icon in the Pages panel deletes the page at once. There is no confirmation, and there is no undo for it. A page was really lost this way while testing.

The icon is also in the middle of the row. Measured at the default panel width (147 px), across the height of a row:

| position in the row | what is there |
|---|---|
| 10 % | the page name, selects the page |
| 25 % | the drag handle |
| **50 %** | **the trash icon, deletes the page** |
| 75 % and 90 % | the gear icon |

So the area that selects a page is about 15 % of the row, and the natural click "I want this page" lands on delete.

### Where it is

`editor/grapesjs/page-panel.ts:275` wires the trash icon to `removePage()`.

A `removePageWithConfirm()` already exists in the same file, line 104, with an "Are you sure?" dialog and a Cancel button. It is only wired to the `cmdRemovePage` command, line 315. There are two ways to delete a page, only one asks, and it is not the one users click.

### Expected

Deleting a page asks for confirmation, whichever way it is triggered. And the click target that selects a page is bigger than the one that destroys it.

britsync07-prog:
Resolves #1839

### Description
Clicking the trash icon in the Pages panel previously deleted the page immediately without confirmation. Additionally, the trash icon was situated in the middle of the row directly next to the page name, leading to accidental deletions when attempting to select a page.

### Changes
- **Confirmation dialog on trash click**: Wires the trash icon in `editor/grapesjs/page-panel.ts` to `removePageWithConfirm()` instead of `removePage()`, prompting the confirmation modal.
- **Safer button layout**: Moved the trash icon to the far right of the row (after clone and settings) so the page selection target has ample space and accidental clicks are avoided.
- **Event propagation**: Added `e.stopPropagation()` on row action buttons (clone, settings, delete) to prevent unintentionally triggering row selection when clicking actions.
- **Single-page guard**: Added a guard in `removePageWithConfirm` to return early if only one page exists.

### Verification & Testing
- ESLint: 0 errors, 0 warnings (`npx eslint editor/grapesjs/page-panel.ts`)
- DOM Unit Tests: 6/6 test suites passed (42/42 tests passing) (`jest --selectProjects dom`)
- Webpack Client Build: Compiled successfully (`pnpm run build:client`)
- TypeScript Server Build: Compiled successfully (`pnpm run build:server`)


lexoyo (approved):
Thanks, this is exactly what #1839 needed

PR #1861, opened 2026-09-17
fix(symbols): unlink selected component

### LINKED ISSUE #1860: MCP symbols_unlink fails with "missing param component"
Calling `symbols_unlink` returns an error: `Can not unlink the component: missing param component`. Seen on Silex Desktop built from `main` on 16 September 2026.

1. Open Silex Desktop and connect an AI agent to its MCP server.
2. Ask the agent to create a website, add a text and turn it into a symbol.
3. Ask the agent to unlink the symbol. The agent gets the error above.

The tool takes no parameter, but the command it runs expects a `component`. It should unlink the selected element.

mikemikimike:
Fixes #1860

Issue: https://github.com/silexlabs/Silex/issues/1860

When the unlink command is invoked without an explicit component, use the currently selected component. Preserve the existing missing-component error when nothing is selected.


grapesjs-plugins/grapesjs-symbols/src/SymbolsCommands.ts
```diff
@@ -88,7 +88,7 @@ export function _removeSymbol(
 
 export function _unlinkSymbolInstance(
   editor: Editor,
-  _: any, { component }: { component: Component },
+  _: any, { component = editor.getSelected() }: { component?: Component },
 ) {
   if(!component) {
     throw new Error('Can not unlink the component: missing param component')
```
lexoyo: Using the selected element is the right idea: an AI agent only sends JSON, so it can not pass a component object. It selects the element with components:select first.

The error is read by the AI agent, so it should say what to do. "missing param component" is not true anymore, because the MCP tool has no such parameter. Could you change it to something like "No element selected. Use components:select first."?

lexoyo (changes requested):
this comment was not releveant

lexoyo (approved):
this comment was not releveant

lexoyo (approved):
Thank you, this fixes the bug
Next time, could you please add a screenshot of your tests (I have to ask to make sure new contributors actually test themselves, not with AI, not rely on me)
Also I'll suggest other issues you can test if you are up to it?
Have a nice day!

lexoyo:
@mikemikimike thank you again for this fix! If you want to continue, #1868 is close to what you did: some MCP tools return a success when nothing was done. It can be done one plugin at a time. Would you like to work on it?

PR #1862, opened 2026-09-18
fix(data-source): synchronize live canvas DOM

### LINKED ISSUE #1845: Removing an attribute does not remove it from the canvas
When you remove an attribute in the Attributes section of the element settings, the attribute stays on the element in the canvas. It only goes away after a reload.

Steps to reproduce, in Silex:

1. Select any element in the editor
2. Open the element settings and go to the Attributes section
3. Click "+" and add an attribute named `title` with the value `hello`
4. Remove that same attribute from the list

Then, in the browser developer tools:

5. Inspect the element inside the editor canvas
6. The attribute `title="hello"` is still on the element

Expected: the attribute is removed from the element in the canvas as soon as you remove it from the list.

Actual: the attribute stays until you reload the page. After a reload the website opens without the attribute, so the saved data is correct and only the canvas is out of date.

Tested on v3.silex.me and on canary.silex.me.

One more thing to check: the HTML content and Visibility condition fields are rendered the same way. Please check if they have the same problem.

@SparshM07 you have worked on the editor recently. Would you like to take this one?

SparshM07:
## Summary

Fixes #1845.

- Synchronize dynamically rendered attributes with the live canvas DOM.
- Remove stale dynamic attributes when they are deleted.
- Restore base component attributes when a dynamic override is removed.
- Restore original component content when dynamic HTML is cleared.
- Verify visibility condition synchronization behavior.
- Add regression tests covering attribute, HTML content, and visibility synchronization.

## Verification

- 56 plugin tests passing
- Full JS build passing
- ESLint passing with 0 errors and 0 warnings


grapesjs-plugins/grapesjs-data-source/src/view/canvas.ts
```diff
@@ -447,6 +493,31 @@ export function restoreOriginalRender(comp: Component) {
...
+            view.el.setAttribute(attrName, String(baseVal))
+          }
+        } else {
+          view.el.removeAttribute(attrName)
+        }
+      })
+      renderedAttributesMap.delete(view.el)
+    }
+    renderedInnerHTMLMap.delete(view.el)
+  }
```
lexoyo, line 539: I think this block may not be needed, because `view.render()` just after rebuilds the attributes from the model. Could you check? If it is needed, maybe the restore code could go in one function, so it is not written twice

lexoyo (commented):
One small thing I noticed:

1. Select an element
2. Add a `class` attribute with a fixed value
3. Remove it

The classes come back, but the element loses its selection outline (`gjs-selected`) until I click it again. Could you have a look?

When it is ready, could you test it in your browser and add a screenshot to the PR?

SparshM07:
Hi @Alex,

Thanks for the detailed review and reproducible scenario. I investigated the gjs-selected issue and addressed the comments.

1. Selection outline fix

The issue was caused by restoring the class attribute directly, which restored only the model classes and overwrote GrapesJS runtime status classes such as gjs-selected.

I changed the class restoration to use GrapesJS' native view.updateClasses()/status handling, so model classes are restored without losing the active editor status classes.

2. restoreOriginalRender()

You're right that the manual attribute restoration there was redundant because view.render() rebuilds the model attributes/classes. I removed that block and now only clear the tracking maps before calling view.render().

3. Hidden components

renderContent() and renderAttributes() now run before the visibility check, so dynamic HTML and attributes are reconciled even while a component is hidden.

4. Tests

Added regression coverage for:

Dynamic class removal while preserving gjs-selected
Dynamic attribute removal on hidden components
Dynamic HTML removal on hidden components
The event-driven COMPONENT_STATE_CHANGED flow without manually calling doRender()

5. Verification

8 test suites, 65/65 tests passing
ESLint: 0 errors, 0 warnings
Plugin build passing
Monorepo build passing
Browser verification completed successfully

I also reproduced the exact scenario you mentioned in the browser. After removing the dynamic class, the original classes were restored and gjs-selected, the selection toolbar, bounding box, and resize handles all remained active without requiring a re-click.
pr_1862_browser_verification

Issue #1856, opened 2026-09-11
Templates are not available with the FTP connector, nor on local file storage


lexoyo:
Templates only work with the GitLab connector. With FTP, and with the local storage of Silex Desktop, the user gets `This storage connector does not support forking external projects`.

Unlike #1855 this never worked, it is not a regression.

How to reproduce:

1. Open https://v3.silex.me/ and log in with FTP
2. Create a new website, click on "from a template"
3. Click "Open in Silex" on any template
4. You get `This storage connector does not support forking external projects`

Why this happens:

- templates are always on gitlab.com, in a public repo
- only the GitLab connector can fork a project
- FTP and local storage cannot fork at all

How to fix it:

* templates are public, so reading one needs no account
* download the repo as an archive and unpack it into a new site, instead of copying file by file

PR #1863, opened 2026-09-19
fix(gitlab): import templates by URL when connected to another GitLab instance

### LINKED ISSUE #1855: Opening a template fails when the user is connected to another GitLab instance
Reported by ceubri on the forum: https://community.silex.me/d/317

A user connected with the gitlab connector but from another instance than gitlab.com - e.g. framagit.org, cannot open any template. They get `Project not found: silex-templates/silex_devdocs-template`. The template does exist, but on gitlab.com, and Silex looks for it on framagit.org.

How to reproduce:

1. Open https://v3.silex.me/ and log in with framagit.org
2. Create a new website, click on "from a template"
3. Click "Open in Silex" on any template
4. You get `Project not found: silex-templates/silex_devdocs-template`

The same steps work when you log in with gitlab.com.

Why this happens:

- templates are always on gitlab.com, in a public repo
- GitLab cannot fork a project of a different instance

How to fix it:

* GitLab connector should fork only if the user is logged in on gitlab.com
* otherwise, create the project with `import_url`, the "Repository by URL" import of the GitLab API
* the destination instance must have this import source enabled. It is on by default on gitlab.com, to be checked on framagit.org

FTP and the local storage will be covered separately in #1856. Templates never worked there at all.

DevShoaib78:
## Related discussion

Fixes #1855, which carries the approach you described there: fork only on gitlab.com, otherwise create the project with the "Repository by URL" import, and say so clearly if that import source is disabled. #1856 (FTP and local storage) is untouched.

## What changed

`server/plugins/GitlabConnector.ts`. `forkWebsite` now checks `isUsingOfficialInstance()` first.

On gitlab.com nothing changes: same path validation, same project lookup, same fork body, same 60 second poll, same messages. The only movement is that the poll loop became `waitForImport`, with the error prefix as a parameter.

On any other instance it calls the new `importWebsite`: read the template's name and `http_url_to_repo` from the gitlab.com public API without a token, then `POST /api/v4/projects/` with `import_url` and `visibility: private` in the user's namespace, then poll for `import_status` the same way. If the instance has the import source disabled, the user gets a message naming the instance and where an administrator enables it, instead of "Project not found".

Two things worth knowing. The user's token never leaves their instance; the template lookup on gitlab.com is anonymous, which is fine because templates are public. And an imported project has no `forked_from_project`, so `getWebsiteMeta` will not show the "forked from" template information that a real fork shows. That is inherent to `import_url`.

The 403 detection comes from GitLab's own source: `lib/api/projects.rb` sets `import_type = 'git'` when `import_url` is present and answers `forbidden!` on `import_source_disabled`. Older versions surfaced it as a validation error instead, so both are handled.

## How to test

`corepack pnpm test` covers this with 11 new tests in `server/plugins/GitlabConnector.test.ts`: the gitlab.com path still forks and never sends `import_url`, another instance imports by URL and never calls `/fork`, the poll walks scheduled to started to finished, and the disabled import source is reported on both the 403 and the older validation error.

I could not run the live check, since I have no account on a self-hosted GitLab. The real test is the one in the issue: log in on v3.silex.me with framagit.org, create a new website, choose "from a template", and confirm the project is created instead of failing with "Project not found". Whether framagit.org has that import source enabled is the open question from the issue; if it does not, the user now gets the explicit message.

`pnpm lint`, `pnpm build:server` and `tsc --noEmit` are clean. `pnpm test` goes from 101 to 112 passing; the two suites that fail here (`FtpConnector`, `DownloadConnector`) fail the same way on a clean checkout because they assert on `/` path separators and I am on Windows.


lexoyo:
Thanks for this, @ceubri and other users will appreciate the fix!

My first feedback before I can make a full code review:
1. There are too many tests. more than we can maintain. Please keep two or three for the important cases and drop the rest
1. There is some code duplication, please double check
1. I need to make sure (because of AI) that you test by hand before you create a PR, would you please add a screenshot of one of our templates cloned into your account on a GitLab instance that is not gitlab.com. For example you can clone this template: https://gitlab.com/silex-templates/silex_devdocs-template

Have a great day

DevShoaib78:
Thanks @lexoyo, all fair points.

1. **Tests:** down to three, the cases that matter. On gitlab.com it still forks, and the request is unchanged. On another instance it creates the project from the template URL instead. And when the instance has the "Repository by URL" import source disabled, the user gets a clear error. Deliberately breaking the fix makes the two other-instance tests fail.
2. **Duplication:** you were right. The import path repeated the fork path: reading the source project, generating the name, creating the project and waiting for it. Now both go through one flow and only the final GitLab call differs (`/fork` vs create-with-`import_url`). The separate `importWebsite` and `waitForImport` helpers are gone, and so is a `@ts-ignore`. The gitlab.com path reads almost exactly as it did before this PR, and the connector diff is about half the size.
3. **Manual test:** I'm setting up a local Silex connected to framagit.org and will post a screenshot of `silex_devdocs-template` opened into my account there as soon as it's done.

`pnpm lint` is clean on the changed files and the type check passes. `pnpm test` gives 104/107: the three failures are in `DownloadConnector` and `FtpConnector` and fail identically without this change (Windows path separators on my machine).

Issue #1866, opened 2026-09-19
MCP: the agent gets wrong names


lexoyo:
1. MCP clients do not accept `:` in names, so `mcp.rs` replaces `:` with `_` in schema properties (line 283). Silex does not know the new names: `og_title` is saved, but Silex reads `og:title`.

   `addCapability` in `grapesjs-ai-capabilities` should refuse an invalid property name in `inputSchema`. The settings schema should use `og_title`, `og_description`, `og_image`, and `settings:set` and `settings:get` map them to the `og:*` keys. Tool names are changed in #1864.
2. The `warnings` sent with each answer (`desktop/src-tauri/scripts/desktop-bridge.js`, lines 65 to 67) name tools that do not exist: `page(action:'select')`, `component(action:'select')`, `selector(action:'select')`, `style(action:'set')`.
3. `selection.component` (same file, line 59) is the `ccid` of the element, but `components_list` and `components_select` use `getId()`. When an element has an `id` attribute, the id in the selection can not be used with `components_select`.

dyk1454683243-sudo:
Taking this issue — all three bullets:

1. Reject invalid schema property names in `addCapability`, use underscore names (`og_title` / `og_description` / `og_image`) in the settings schema, and map them in `settings:set` / `settings:get`.
2. Update the desktop-bridge warning strings to the real MCP tool names after #1864.
3. Make selection report a component id that `components_select` accepts (prefer `getId()` when present).

#1868 (no-op success / fonts) stays out of scope — separate PR.

PR #1870, opened 2026-09-19
fix(mcp): reject silent no-ops and match font family case-insensitively

### LINKED ISSUE #1868: MCP tools return a success when nothing was done
These tools return a success when their target does not exist:

- `classes_remove`: class not on the element
- `styles_remove`: property not set
- `css-var_remove`, `css-var_rename`: unknown variable (and `css-var_rename` overwrites an existing `newName`)
- `css-var_set`: the new `type` of an existing variable is ignored
- `data-source_remove-state`: unknown `stateId`
- `history_undo`, `history_redo`: nothing to undo or redo

Each one should return an error that says what to do, and gives the valid values when there are some, for example "Class "x" is not on the selected element. Its classes are: card, hero."

Also: `fonts_remove` says `Font "roboto" not installed` when `Roboto` is installed. The name is compared with the case (`grapesjs-fonts/src/commands.js`, line 93), while `fonts_install` ignores the case.

dyk1454683243-sudo:
## Related discussion

Fixes #1868.

This does **not** address #1866 (wrong MCP names / selection id mismatch). That stays a separate PR.

## What changed

Mutating MCP commands now throw when the requested postcondition did not happen, instead of returning success. Desktop MCP already surfaces thrown `runCommand` errors as tool errors (`mcp.rs`), so no Tauri glue change was needed.

- `classes:remove` — error if the class is not on the selected element, listing its current classes
- `styles:remove` — error if the property is not set, listing current style keys
- `css-var:remove` — error if the variable does not exist, listing current variables
- `css-var:rename` — error if `oldName` is unknown, or if `newName` would overwrite an existing variable
- `css-var:set` — error if the requested `type` would be ignored on an existing variable
- `data-source:remove-state` — error if `stateId` is unknown, listing current state ids
- `history:undo` / `history:redo` — error when the history stack is empty
- `fonts:remove` — matches family names case-insensitively (same as `fonts:install`), so `fonts_remove("roboto")` removes installed `Roboto`; still errors with the installed list when the family is truly missing

## How to test

Plugin unit tests (reject + happy paths), all passing locally:

```sh
pnpm --filter @silexlabs/grapesjs-fonts test
# 5/5 passed — wrong-case remove of Roboto via "roboto", missing family lists installed fonts

pnpm --filter @silexlabs/grapesjs-css-variables test
# 8/8 passed — missing/overwrite/type-mismatch reject, valid rename/set/remove names

pnpm --filter @silexlabs/grapesjs-advanced-selector test -- src/commands.test.ts
# 4/4 passed — missing property lists current keys; set property accepted

pnpm --filter @silexlabs/grapesjs-data-source test -- src/commands.test.ts
# 4/4 passed — unknown stateId lists current ids; existing state removes

pnpm test -- editor/grapesjs/core-commands.test.ts
# 6/6 passed — classes:remove reject/happy; history undo/redo empty and after a change
```

In Silex Desktop, with an element selected that has classes `card` and `hero`:

1. Call `classes_remove` with `name: "x"` — expect an error listing `card, hero`, not success
2. Call `classes_remove` with `name: "card"` — class is removed
3. Install Roboto, then call `fonts_remove` with `family: "roboto"` — Roboto is removed


dyk1454683243-sudo:

Issue #1864, opened 2026-09-19
MCP: one tool per noun with an action parameter, for small local models


lexoyo:
Goal: Silex Desktop must work with small local models (7B to 8B). They get worse with more than about 15 tools. Today the MCP has more than 50 tools.

Rules for the new surface:

1. One tool per noun (`website`, `page`, `component`, `style`, `symbol`, `data`, `device`...) with an `action` parameter. All actions are listed in the `enum` of the schema.
2. Tools act on the selection. The only way to choose a target is a `select` action. An id is only used for a second object (the parent in `move`, the symbol in `symbol create`). So `components_remove` loses its `id`, `components_move` moves the selected element, `pages_delete` and `pages_rename` act on the selected page, and `website` delete, rename and duplicate act on the open website (open is the `select` of websites).
3. When several elements are selected, the action is done on all of them.
4. Settings take a `scope` parameter, `site` or `page` (the selected page), instead of `page` being a name or any other value.
5. Every parameter is declared in the schema, with one type.
6. Every error says what is wrong and what to do next, and lists the valid values when there are some.
7. Every answer shows the selection. The `warnings` (next step) are only in the answers of a tool that fails.
8. What a tool creates (an element, a page, a symbol instance) becomes the selection, and the description of the tool says it.
9. All tools use the same verb for the same intent: `list`, `select`, `add`, `update`, `remove`, `rename`.
10. The instructions of the MCP server (`mcp.rs`, around line 725) are rewritten in simple English, short, with an example of 4 calls in a row.
11. The description of each tool has one line per action and a full example call, for example: `select: choose the element the next actions apply to. Example: {"action":"select","id":"i3k"}`.
12. Before and after the change: measure the size of the tool list in tokens, and test with a small local model that builds a website of 3 pages.

PR #1871, opened 2026-09-19
fix(mcp): map settings keys, warning names, and selection ids

### LINKED ISSUE #1866: MCP: the agent gets wrong names

dyk1454683243-sudo:
## Related discussion

Fixes #1866.

Taking all three bullets (commented on the issue). Out of scope: #1868 (no-op success / fonts) — that stays a separate PR.

## What changed

1. **Schema property underscores.** MCP clients reject `:` in property names (`^[a-zA-Z0-9_.-]{1,64}`). `addCapability` now refuses invalid `inputSchema` property names. The settings schema uses `og_title` / `og_description` / `og_image`. `settings:set` maps those to stored `og:*` keys; `settings:get` maps them back. Leftover `og_*` keys from the previous rewrite bug are dropped on set.

2. **Stale warning tool names.** #1864 (one tool per noun + `action`) is still an open redesign and is not on `main`. Current MCP tool names on `main` are the capability ids after `:` → `_` in `mcp.rs`. The desktop-bridge warnings now cite those real names only: `pages_select`, `components_select`, `selector_set`, `styles_set` (there is no `selector:select` — the selector is set with `selector:set`).

3. **Selection id mismatch.** Smallest correct fix: `getSelectionState` reports `getId()` when present, and falls back to `ccid`. `components_list` / `components_select` already look up `getId()` (`attrs.id || ccid`), so the id in `selection.component` can be passed to `components_select`.

## How to test

```sh
pnpm --filter @silexlabs/grapesjs-ai-capabilities run build
pnpm --filter @silexlabs/grapesjs-ai-capabilities run test
pnpm exec jest --selectProjects dom --testPathPatterns='settings-mcp|components-select'
node desktop/src-tauri/scripts/desktop-bridge.test.js
```

In Silex Desktop, after opening a project:

1. `settings_set` with `{ "og_title": "Hello" }` — site settings store `og:title`; `settings_get` returns `og_title`.
2. A tool response with no selection should warn to use `pages_select` / `components_select` / `selector_set` / `styles_set`, not `page(action:'select')`.
3. Select an element that has an HTML `id`, then pass `selection.component` to `components_select` — it should select that element.


dyk1454683243-sudo:

Issue #1725, opened 2026-06-11
Publication zips from the Download connector are never deleted and fill the server disk


lexoyo:
## The problem

When a user publishes their website with the **"Download zip file"** connector, Silex creates the zip in the OS temp folder — and never deletes it.

On a long-running server, these zip files (40-60 MB each for a typical site) accumulate forever. On v3.silex.me this filled the whole disk: we found **18 GB of zip files** (~380 of them) in the container's `/tmp`, which broke deployments on the server.

## Where to look

Everything happens in [`src/ts/plugins/server/DownloadConnector.ts`](https://github.com/silexlabs/silex-lib/blob/main/src/ts/plugins/server/DownloadConnector.ts) (in the `silex-lib` repo):

1. `startPublishingInBackground()` creates the zip at `${tmpdir()}/${fileName}` ([line 84](https://github.com/silexlabs/silex-lib/blob/main/src/ts/plugins/server/DownloadConnector.ts#L84))
2. The constructor registers the route `GET /download/:tmpZipFile` which serves the file with `res.sendFile()` ([line 31](https://github.com/silexlabs/silex-lib/blob/main/src/ts/plugins/server/DownloadConnector.ts#L31))
3. Nothing ever deletes the file

## Suggested fix

Two complementary parts (the first one is enough to close this issue):

1. **Delete the zip after it is downloaded**: in the `/download/:tmpZipFile` route, delete the file in the `res.sendFile()` callback once the transfer succeeded (`fs.unlink`).
2. *(Bonus)* **Sweep leftovers on startup or periodically**: files matching the zip naming pattern (`<websiteId>-<timestamp>-<random>.zip`) older than e.g. 24h can be deleted safely — this covers zips that were created but never downloaded.

Notes:
- Keep in mind the user may click the download link more than once: deleting after the *first* successful download is the accepted trade-off (the link is presented as a one-time download), but mention it in your PR if you choose otherwise.
- The filename is user-controlled in the route (`:tmpZipFile`) — while you are at it, reject path separators / `..` in the parameter to avoid serving arbitrary files from the temp dir.

## How to test locally

1. Clone [silex-lib](https://github.com/silexlabs/silex-lib), run `yarn && yarn build && yarn start`
2. Open Silex, publish any website choosing "Download zip file"
3. Look at your OS temp dir (`/tmp` on Linux/macOS): a `<websiteId>-<timestamp>-<random>.zip` appears
[...]

PR #1873, opened 2026-09-19
fix(download-connector): sweep never-downloaded zips from tmp

### LINKED ISSUE #1729: Download connector: sweep never-downloaded zips from temp dir
Follow-up of #1725 / silexlabs/silex-lib#24.

#24 deletes a zip after it's downloaded. But zips that are never downloaded stay
forever — that's what filled `/tmp` with 18 GB on v3.silex.me.

## What to do

In `DownloadConnector.ts` (silex-lib), sweep old zips from `os.tmpdir()`:
- run once at startup, then on a daily `setInterval` (`.unref()` it)
- match only the generated pattern `/-\d{13}-[a-z0-9]+\.zip$/`
- delete matches older than 24h (mtime)

## Definition of done

- [ ] Matching zips older than 24h are removed at startup and once a day
- [ ] Non-matching files are never touched; recent zips are kept
- [ ] The daily timer does not block the process (`.unref()`)
- [ ] Unit test: old deleted, recent kept, non-matching ignored

dyk1454683243-sudo:
## Related discussion

Fixes #1729.

Follow-up of #1725 (unlink after a successful download). This covers zips that are never downloaded and would otherwise stay in `os.tmpdir()`.

Implemented in the Silex monorepo `server/plugins/DownloadConnector.ts` (not archived silex-lib).

## What changed

On `ServerEvent.STARTUP_END`:

- Sweep `os.tmpdir()` once
- Schedule a daily `setInterval` and `.unref()` it so the timer does not keep the process alive
- Delete only files whose names match the generated pattern (`websiteId` + 13-digit `Date.now()` + alphanumeric random + `.zip`) and whose mtime is older than 24h
- Non-matching names are never stat'd or unlinked; recent matching zips are kept

## How to test

```sh
pnpm test -- server/plugins/DownloadConnector.test.ts
```

Local run (Node 24): **5/5 passed**, eslint clean on the two touched files.

- `deletes matching zips older than 24h and leaves recent and non-matching files`
- `sweeps once at startup and schedules a daily unref timer`
- `sweeps again when the daily timer fires`
- existing download-route unlink / path-traversal cases still pass


dyk1454683243-sudo:

PR #1876, opened 2026-09-19
fix(mcp): select created content and reject illegal moves

### LINKED ISSUE #1867: MCP: tools that add or move content do not give the agent what it needs
After these tools, the agent can not work on what was added or moved, or can not use the tool at all. What a tool creates (an element, a page) should become the selection, and the description of the tool should say it.

- `components_move` (`editor/grapesjs/core-commands.ts`, lines 87 and 88) appends a clone and removes the original, so the element gets a new id. GrapesJS can move a component and keep its id (`component.move(dest, { at })`). Also: moving an element into one of its own children removes it, and a `position` that is not a number is ignored.
- `blocks_add` (`core-commands.ts`, line 46) does not select the new element. It can also add the block inside an element that can not have children. Expected: the new element is selected, and an error when the selected element can not have children. When nothing is selected, the block goes into the body, like for a user.
- `pages_add` opens the "new page" dialog (`editor/grapesjs/page-panel.ts`, line 54). Expected: a `name` parameter, and no dialog from the MCP. The new page is already selected, the description of the tool should say it. The dialog stays for the button in the editor.
- `symbols_create` has no input schema, and the command expects objects (`symbol`, `pos`, `target`) that can not be sent as JSON, so it always fails. Expected, like `blocks_add`: the instance is added inside the selected element and becomes the selection, and the tool takes a `symbolId`. `symbols_list` must then return the id of each symbol (today it returns full component objects and no id). The plugin is published on npm, so the command can keep its parameters, and the MCP tool can use a new command.

GrapesJS already checks if an element can go somewhere: `editor.Components.canMove(target, source)`. It can be used in `components_move` and `blocks_add`, and its `reason` returned as the error.

dyk1454683243-sudo:
## Related discussion

https://github.com/silexlabs/Silex/issues/1867

## What changed

Fixes #1867.

After MCP tools that add or move content, the agent now gets a usable selection (and a clear error when the action is illegal), so it can keep editing what was just created or moved.

- **`components:move`**: uses `component.move()` so the id is unchanged; `canMove` rejects illegal destinations (including a parent into its own child) and returns the `canMove` reason; a non-number `position` is an error.
- **`blocks:add`**: selects the new element (body when nothing is selected); `canMove` rejects a parent that cannot have children.
- **`pages:add`**: MCP `{ name }` creates and selects the page with no new-page dialog. The editor button / keymap still open the dialog.
- **`symbols:list`**: returns `{ id, name, instances }` instead of opaque component objects.
- **`symbols:create`**: MCP capability now runs a new `symbols:create-instance` command that takes `symbolId`, inserts under the selection, and selects the instance. The published `symbols:create` command signature is unchanged.

Capability `description` / `inputSchema` strings say the new element or page is selected.

## How to test

Automated, all passing locally:

```sh
pnpm test -- editor/grapesjs/core-commands.test.ts editor/grapesjs/page-panel-mcp.test.ts
# 10/10 passed — move keeps id, illegal move / non-number position / non-container parent error, pages:add by name skips dialog

pnpm test
# 114/114 passed (full editor + server + common suite)

pnpm --filter @silexlabs/grapesjs-symbols run test
# 33/33 passed — list returns ids; create-from-id selects the instance; canMove reject

pnpm lint
pnpm --filter @silexlabs/grapesjs-symbols run lint
# both clean
```

Manual MCP (desktop, website open):

1. `blocks_add` with a `blockId` from `blocks_list` — response `selection.component` is the new element. Adding into a text/image (non-container) returns a `canMove` error.
2. `components_move` — the same `id` still works; moving a parent into a child errors with `canMove` reason.
3. `pages_add` with `{ "name": "About" }` — no dialog, `pages_list` shows About selected.
4. `symbols_list` returns ids; `symbols_create` with `{ "symbolId": "..." }` inserts under the selection and selects the instance.

The editor **+** button for a new page still opens the dialog.


dyk1454683243-sudo:

PR #1877, opened 2026-09-20
fix(editor): reset isSaving after a store with no user

### LINKED ISSUE #1832: Saving stops working after a save while logged out
If a save happens while the editor has no user, saving stops working for the rest of the session. Nothing is shown, and logging back in does not help. Only reloading the page does.

To reproduce, from the console:

```js
editor.getModel().set("user", null)   // what Log out does
await editor.store()                  // nothing is saved
editor.getModel().set("user", /* a valid user */)
await editor.store()                  // still nothing, and never again
```

In `storage.ts`, `doStore` sets `isSaving = true` before reading the user, and only sets it back to `false` when there is one. When there is not, `isSaving` stays `true` and every later save waits for a `storage:end:store` that never comes.

The 401 case is handled, the catch resets `isSaving`. It is the path without an exception that is missed.

dyk1454683243-sudo:
## Related discussion

Fixes #1832

## What changed

`doStore` set `isSaving = true` before `getCurrentUser` and only cleared it after a successful save. When there was no user, `isSaving` stayed true and every later store waited for a `storage:end:store` that never fired.

The 401 catch already reset the flag. This wraps the save path in `try/finally` so the empty-user path (no exception) also clears it. Publication-pending delay and `lastPendingSaving` cancel/queue behavior are unchanged.

The grapesjs in-flight dirty-loss case from the issue comment is a different cause (GrapesJS does not call `store()` for edits made during a slow save) and is left for a follow-up.

## How to test

Unit tests in `editor/grapesjs/storage.test.ts` cover:

- store with no user, then store with a user → `websiteSave` runs
- 401 then a later store → `websiteSave` runs again
- publication pending → save waits for `silex:publish:end`

Verified locally: `pnpm lint`, `pnpm test` (107 passed), `pnpm build`.

Console repro from the issue (after login, in the editor):

```js
editor.getModel().set("user", null)
await editor.store()
editor.getModel().set("user", /* a valid user */)
await editor.store()  // should save again, without reload
```


dyk1454683243-sudo:

Issue #1869, opened 2026-09-19
MCP website and take_screenshot: success when the action failed


lexoyo:
In `desktop/src-tauri/src/mcp.rs`:

- `website` open and create return a success when the editor tools failed to load (only a log line), and open returns a success for an id that does not exist
- `website` delete goes back to the dashboard and closes the open website, even when another website was deleted. It should only go back to the dashboard when the deleted website is the open one
- error messages of delete, rename and duplicate do not give the HTTP status
- `take_screenshot`: an unknown `target` is used as `ui` without an error; when saving the file fails, the result is still a success
- "No project open", "website_id is required", "name is required" and "Tool not found" are sent as protocol errors, some clients only show "internal error". They should be tool errors, like the others.

dyk1454683243-sudo:
Taking this: I will make `website` open/create/delete and `take_screenshot` report failure when the action did not succeed (including open of an unknown id, capabilities not loading after open/create, delete of a non-open site forcing the dashboard, unknown screenshot target, and a failed screenshot save), and turn the user-facing validation messages (`website_id is required`, `name is required`, `No project open`, `Tool not found`) into tool errors (`is_error`) so clients show the text instead of an opaque protocol error.

No open PR claims Fixes #1869. Separate branch from current `main` — not stacked on #1870–#1874.

PR #1878, opened 2026-09-20
fix(mcp): await async commands, page fonts, and follow publish

### LINKED ISSUE #1865: MCP: async commands return an empty success, and publish can not be followed
When an editor command is async, the MCP tool returns `{"selection": {...}}` right away. The agent never gets the result, and never gets the error. Seen with `fonts_available` (the font list never comes back) and `publish`.

When it works, `fonts_available` will return all Google Fonts (about 1700). It should return at most 20 fonts, and say how many are left, so the agent uses `search`.

A publication can take several minutes, more than the time limit of an MCP call. So `publish` takes an `action` parameter:
- `start` starts the publication and answers at once: `{status, message: "Publication started. Call publish with action status to follow it."}`
- `status` returns the state of the publication: pending, success with the URL, or error with what to do (for example "No hosting is set for this website. Ask the user to choose where to publish in the Publish dialog, then retry."). It also returns the errors of the job, and the last lines of its logs.
- while the publication runs, `status` answers: "Still publishing. Call publish with action status again. Do not tell the user it is done."

There is only one publication at a time in Silex Desktop, so no job id is needed.

Today the answer of a tool that fails has no `selection` (`mcp.rs`, line 336), so the agent does not see which element is selected. All answers should have the `selection`, the answers of a tool that fails too. The `warnings` should only be in the answers of a tool that fails: today they are in every answer, even when they are not about the tool that was called.

dyk1454683243-sudo:
## Related discussion

https://github.com/silexlabs/Silex/issues/1865

## What changed

**Fixes #1865.**

Separate branch from current `main`. Not stacked on #1870–#1876. Does not reopen #1868 / #1866 / #1869 / #1867.

- **Async commands.** The dynamic-tool JS bridge is now `async` and `await`s a Promise from `runCommand` (or `runPublish`). The agent gets the real result or error instead of `{"selection":{...}}` from stringifying a Promise.
- **`fonts_available`.** At most 20 fonts, plus `remaining` and a hint to use `search`. Capability description says so.
- **`publish`.** `action=start` starts publication and answers at once: `{status, message: "Publication started. Call publish with action status to follow it."}`. `action=status` returns pending / success+URL / error with the next step, job errors, and the last log lines. While it runs: `"Still publishing. Call publish with action status again. Do not tell the user it is done."` One publication at a time, no job id. Desktop-side job state lives on the webview bridge and reads `PublicationManager`.
- **Response shape.** Failures include `selection` (the old catch path did not). `warnings` are stripped from successful answers.

## How to test

```sh
node desktop/src-tauri/scripts/desktop-bridge.test.js
# 12/12 passed
```

Rust helpers in `desktop/src-tauri/src/mcp.rs` (`dynamic_tool_answers`): await in the eval JS, font page + remaining, warnings only on failure, selection on eval errors, publish pending/success are not tool errors.

```sh
cargo test -p silex-desktop dynamic_tool
# needs GTK/WebKitGTK (Tauri prerequisites). Not run in this environment — gdk-sys pkg-config.
```

Manual MCP in Silex Desktop (`http://localhost:6807/mcp`), website open:

1. `fonts_available` — a font list comes back (not empty success). At most 20 fonts, `remaining` if there are more. `search` narrows the list.
2. `fonts_available` with a bad `api_key` / network — error text, plus `selection`.
3. `publish` with `action: "start"` — immediate `{status:"pending", message:"Publication started. Call publish with action status to follow it."}`.
4. `publish` with `action: "status"` while it runs — `"Still publishing. ... Do not tell the user it is done."`
5. After success — `status:"success"` and the published URL.
6. No hosting chosen — status error: `"No hosting is set for this website. Ask the user to choose where to publish in the Publish dialog, then retry."` plus job errors / last log lines when there are some.
7. A successful tool answer has `selection` and no `warnings`. A failed one has both.


dyk1454683243-sudo:

PR #1879, opened 2026-09-20
fix(data-source): do not console.error on empty selection

### LINKED ISSUE #1842: Console noise in the editor: an error on a normal state, and two logs for services that are not involved
Three sources of console noise in the editor, from the worst to the most harmless.

### 1. `console.error` in bursts, on a state that is normal

`console.error("selected and editor are required", null, null)` is printed 16 times for a single interaction, just opening the Pages panel.

It comes from `grapesjs-plugins/grapesjs-data-source/src/view/state-editor.ts:178`, inside a getter, so it is called in a loop. The `throw` on the next line was commented out on purpose, but the `console.error` severity stayed.

"Nothing is selected" is a normal state, not an error. The real errors are lost in the noise.

### 2. `Gitlab url:` (empty) during a zip export

A connector that has nothing to do with the job speaks up.

### 3. A 404 on `/api/onboarding` and "Error while fetching user from brevo" on every start

Brevo is not configured on a developer machine, and it never will be, so this is permanent noise.

### Expected

No error for a state that is normal, no error for a service that is not configured, and no log from a connector that is not involved.

dyk1454683243-sudo:
## Related discussion

https://github.com/silexlabs/Silex/issues/1842

## What changed

**Fixes #1842.**

Separate branch from current `main`. Not stacked on #1870–#1878.

Empty selection (or an unwired editor) is a normal state — opening the Pages panel is the reported case. `state-editor` no longer treats it as a failure:

- `get data()` returns `[]` with no `console.error` (the commented `throw` is still not restored).
- `render()` returns an empty template instead of `console.error` plus a red “Error rendering…” block.

Real token/expression failures still log at `console.error` (e.g. `Error while getting token from id …`).

### Follow-ups (not in this PR)

Issue #1842 also lists two other console-noise sources. Their call sites are not local to `grapesjs-data-source`:

1. **Empty `Gitlab url:` during zip export** — `console.info('Gitlab url: ', url)` in `editor/grapesjs/PublicationManager.ts` after every publish, including zip. Gating or dropping that log belongs in the publication manager, not a connector.
2. **`/api/onboarding` 404 + “Error while fetching user from brevo”** — the client plugin always POSTs `/api/onboarding`; the server plugin only registers the route when NocoDB env vars are set. Silencing the `console.error` would still leave the browser 404. The real fix is not calling the API when onboarding is off.

## How to test

```sh
pnpm --filter @silexlabs/grapesjs-data-source run test -- src/view/state-editor.test.ts
pnpm --filter @silexlabs/grapesjs-data-source run lint
```

Verified locally on this branch:

- `state-editor.test.ts` — 3/3 passed (empty selection is silent; broken token still `console.error`s)
- full `@silexlabs/grapesjs-data-source` suite — 51/51 passed
- `pnpm --filter @silexlabs/grapesjs-data-source run lint` — clean

### Pages-panel console repro (manual)

1. Open the editor, leave nothing selected (click the canvas background if needed).
2. Open the browser console and clear it.
3. Open the **Pages** panel.
4. Expect **no** `console.error('selected and editor are required', …)` and no red “Error rendering state-editor…” block.
5. Select an element and use a data-source expression with a bad token — `Error while getting token from id …` may still appear at error severity.


dyk1454683243-sudo:

